Issue #63, opened 2018-10-05
Helm support?


vosmith:
You all have probably already discussed this at some point but I could not find any history of it.

Have you already decided not to use Helm for creating your deployments? If it isn't decided, would you be interested in a PR?

joel-bluedata:
Good question I think. Let me cover a couple of different things to make sure I am addressing what you are asking.

For the virtual clusters deployed by KubeDirector, we don't have plans to insert any intermediary like Helm into the process. E.g. we wouldn't try to translate KubeDirectorApp+KubeDIrectorCluster specs into a Helm chart. Currently we don't see an indirection like that as being a net positive, especially given the way that KD will continue to intervene in ongoing management of the virtual cluster and its interaction with other custom resources we may introduce down the road to represent things like data sources, cluster pipelines, etc.

For KubeDirector itself, Helm deployment is a possibility. We have had some discussion about KD lifecycle management (deployment, upgrade, uninstall) and whether that should be via Helm, OLM, some custom solution, or a combination. No decisions yet though. Community suggestions or examples of such solutions for KD could be helpful.

vosmith:
Awesome, thanks.  The second use case you mentioned for deploying KubeDirector itself is what I was thinking of.  I'd like to try and see if I can make a helm chart as a POC.  I'll keep you posted!

123tap:
vosmith, i took the liberty of assigning this PR to you for now.
please unassign if necessary.

PR #71, opened 2018-10-05
[WIP] Added initial helm chart

vosmith:
This commit adds initial support for installing the KubeDirector
Deployment via Helm.  Values are provided to override the number of
replicas, the CRD names, image name/tag, and the service account names.
This chart is configured to pull from bluek8s/kubedirector:unstable from
the master branch and should be updated in other release branches.

Related to #63

### DIFF
--- a/chart/.helmignore
+++ b/chart/.helmignore
@@ -1,0 +1,21 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*~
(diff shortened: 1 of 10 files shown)


joel-bluedata:
Thanks @vosmith; we'll get some folks to take it for a test drive.

joel-bluedata:
Hey @vosmith, just checking back in to say we'll be looking at this more now that we have some of the other fundamentals taken care of.

One thing that I'm wondering about is how we can best avoid duplication of information. For example when the schema for KubeDirectorApp gets updated, we don't want to be required to carefully make equivalent changes to both deploy/kubedirector/crd-app.yaml and also chart/kubedirector/templates/crd-app.yaml . I think we would need a way to generate one from the other, or generate both from some common source... something like that. (And KubeDirectorApp isn't the only instance of this concern.)

joel-bluedata:
We can reopen this when we get more feedback from @vosmith or someone else picks it up. I do think the main blocking issue is just the duplication of data structures mentioned above.

PR #78, opened 2018-10-09
add support for timed running of execCommand() function

123tap:
Joel, was this what you had in mind?

### DIFF
--- a/pkg/executor/guest.go
+++ b/pkg/executor/guest.go
@@ -31,7 +31,8 @@ import (
 	"k8s.io/client-go/util/exec"
 )
 
-const DEFAULT_CMD_TIMEOUT_IN_SECONDS = 60
+// default to 15 minute timeout
+const DEFAULT_CMD_TIMEOUT_IN_SECONDS = 900
 
 // IsFileExists probes whether the given pod's filesystem contains something
 // at the indicated filepath.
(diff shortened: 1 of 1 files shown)


joel-bluedata:
Seems like the sort of thing yep. (I haven't looked at this change in detail yet.)

pkg/executor/guest.go
```diff
@@ -30,6 +31,8 @@ import (
 	"k8s.io/client-go/util/exec"
 )
 
+const DEFAULT_CMD_TIMEOUT_IN_SECONDS = 60
```
joel-bluedata: I think our main use of timeouts will be for enforcing "don't run forever", so the default should be quite large. (Especially if we're not yet providing a way to change the default.) Like, 15 minutes or something else hilariously big.

pkg/executor/guest.go
```diff
@@ -242,14 +278,56 @@ func execCommand(
...
+	timer := time.NewTimer(timeInMilliSeconds)
+	defer timer.Stop()
+
+	// Wait for command completion, or timeout
+	select {
+	case execErr := <-errch:
+		close(errch)
+		return execErr
+	case <-timer.C:
+		close(errch)
```
joel-bluedata: But we'll still have a goroutine out there running the command. Hmm. I wonder if there's a way to get a handle on the underlying connection to close it.

joel-bluedata:
Closing this as dormant.

Issue #80, opened 2018-10-09
"make redeploy" can fail if done immediately after "make deploy"


joel-bluedata:
It can be the case that the KD pod is up but the actual KD container is not quite up yet.

Probably "make deploy" should wait a bit checking for the container to appear before it exits, and/or "make redeploy" should do the same check-and-wait before it tries to do the injections. (We don't want to wait forever though.)

PR #81, opened 2018-10-09
additional validation for app & cluster CRs

joel-bluedata:
Cf. issue #9.

### DIFF
--- a/pkg/validator/app.go
+++ b/pkg/validator/app.go
@@ -27,6 +27,51 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// validateUniqueness checks the lists of roles and service IDs for duplicates.
+func validateUniqueness(
+	appCR *kdv1.KubeDirectorApp,
+	allRoleIDs []string,
+	allServiceIDs []string,
+) string {
+
+	var errorMessages []string
+	if !shared.ListIsUnique(allRoleIDs) {
+		errorMessages = append(errorMessages, nonUniqueRoleID)
+	}
+	if !shared.ListIsUnique(allServiceIDs) {
+		errorMessages = append(errorMessages, nonUniqueServiceID)
+	}
+
+	if len(errorMessages) == 0 {
+		return ""
+	}
+	return strings.Join(errorMessages, "\n")
+}
+
+// validateRefUniqueness checks the lists of role references for duplicates.
+func validateRefUniqueness(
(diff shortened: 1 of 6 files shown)


pkg/validator/app.go
```diff
@@ -82,9 +114,44 @@ func validateSelectedRoles(
...
+// in the schema. Currently this just means checking that the role must
+// specify an image if there is no top-level default image.
+func validateRoles(
+	appCR *kdv1.KubeDirectorApp,
+) string {
+
+	for _, role := range appCR.Spec.NodeRoles {
+		if role.Image.RepoTag == "" {
+			if appCR.Spec.Image.RepoTag == "" {
+				return noDefaultImage
```
123tap: to be consistent with the other uniqueness validating functions,
you may want to use the "var errorMessages []string" and "return strings.Join(errorMessages, "\n")"
construct eventhough there will only ever be at most 1 string in the string array.

joel-bluedata: It might also be the right time for me to introduce a little more infrastructure around the validation functions. I'm not sure I want to go hog wild with that just yet, but at least: if each of those functions is eseentially going to generate a slice of error messages, then that's what each such function should return. The caller can finally compile all of those into a single string (or whatever it wants to do).

Let me see how much that will churn things.

joel-bluedata:
OK, I made a change to handle the accumulation of error messages more cleanly.

Also:

Previously, "make deploy" would create the example app CRs before deploying KD. Since KD is supposed to validate app CRs now, that's not great. I moved the app CR creation until after KD deployment.

This involves waiting for KD to be up. (Or at least, waiting for "kubectl logs" to succeed for that container.) So inadvertently that change should also fix issue #80.

pkg/validator/app.go
```diff
@@ -27,24 +27,63 @@ import (
...
+) []string {
+
+	if !shared.ListIsUnique(allRoleIDs) {
+		valErrors = append(valErrors, nonUniqueRoleID)
+	}
+	if !shared.ListIsUnique(allServiceIDs) {
+		valErrors = append(valErrors, nonUniqueServiceID)
+	}
+	return valErrors
+}
```
swamibluedata: why not mutate by removing the duplicates

joel-bluedata: Just because the ID is duplicate doesn't mean that the whole service or role object is duplicate. If e.g. there are two different role objects with the ID "controller", we don't know which one should be correct, so I think we just have to reject the whole thing.

Issue #40, opened 2018-10-02
in syncCluster when re-trying status updates, use backoff


joel-bluedata:

joel-bluedata:
Going to do this as a quick palate cleanser. I think the status-writeback loop also needs to be modified to bail out if it detects that the cluster CR has been deleted (or deletion-timestamped).

PR #88, opened 2018-10-11
improvements for status-write-retry

joel-bluedata:
issue #40 

- Add backoff for retries.

- Bail out if object is going away or has gone away.

### DIFF
--- a/pkg/reconciler/cluster.go
+++ b/pkg/reconciler/cluster.go
@@ -21,9 +21,11 @@ import (
 	kdv1 "github.com/bluek8s/kubedirector/pkg/apis/kubedirector.bluedata.io/v1alpha1"
 	"github.com/bluek8s/kubedirector/pkg/catalog"
 	"github.com/bluek8s/kubedirector/pkg/executor"
+	"github.com/bluek8s/kubedirector/pkg/observer"
 	"github.com/bluek8s/kubedirector/pkg/shared"
 	"github.com/google/uuid"
 	"github.com/operator-framework/operator-sdk/pkg/sdk"
+	"k8s.io/apimachinery/pkg/api/errors"
 )
 
 // syncCluster runs the reconciliation logic. It is invoked because of a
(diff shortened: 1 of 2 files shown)


pkg/reconciler/cluster.go
```diff
@@ -57,16 +59,43 @@ func syncCluster(
 		if !reflect.DeepEqual(cr.Status, oldStatus) {
 			// Write back the status. Don't exit this handler until we
 			// succeed (will block other handlers for this resource).
```
123tap: this could conceivably never exit. is that ok?

joel-bluedata: Yeah, or at least in the sense that I'm not sure there's much else we can reasonably do here... if we don't update the status on the object then there's no way we can correctly manage it in the future.

A couple of comments on that situation:
* This function not-returning will only block future handling of this specific cluster object. Other objects can still be handled.
* It would be possible (with the change in this PR) to get out of that situation by deleting the object.

There are other ways to handle this situation without tying up a goroutine, and if these seem attractive/necessary we could create an issue to pursue them. For example:
* We could set ourselves a reminder of "don't allow any more changes to this object", save the desired status object (in memory), and schedule a periodic timer to kick off a retry of the status update. But I'm not sure this is notably different from using time.Sleep.
* Or we could (perhaps after some retries) set ourselves a reminder of "don't allow any more changes to this object" and then just mark the object as permanently corrupted.

pkg/reconciler/cluster.go
```diff
@@ -57,16 +59,43 @@ func syncCluster(
...
+				currentCluster, currentClusterErr := observer.GetCluster(
+					cr.Namespace,
+					cr.Name,
+				)
+				if currentClusterErr != nil {
+					if errors.IsNotFound(currentClusterErr) {
+						return
+					}
+				}
+				if currentCluster.DeletionTimestamp != nil {
```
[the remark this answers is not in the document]
joel-bluedata: 👍

pkg/reconciler/cluster.go
```diff
@@ -57,16 +59,44 @@ func syncCluster(
...
+				shared.LogWarnf(
+					cr,
+					"failed to update status: %v",
+					updateErr,
+				)
+				shared.LogWarnf(
+					cr,
+					"trying status update again in %v",
+					wait,
+				)
```
swamibluedata: combine these two messages

PR #104, opened 2018-11-08
Added support for running Spark on Digital Ocean k8s

jeffloyd:
Added YAML file for appropriate settings of running Spark with bluek8s on Digital Ocean's Kubernetes


deploy/example_clusters/cr-cluster-spark221e2.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: "spark-instance"
 spec:
   app: spark221e2
+  serviceType: LoadBalancer
```
joel-bluedata: I don't think we want to make this change; these "non-suffixed" YAML files are intended more for local k8s deployments that use NodePort.

joel-bluedata:
Thanks for the contribution @jeffloyd. :-)  So the difference between this example and the GKE example is just the absence of persistent storage... is that intentional?

Also, would you recommend any other notes about deploying on Digital Ocean that would be interesting? As in comparison to the level of detail in our gke-notes.md doc about GKE deployment.

Our intention in the near future is to move the service type preference and/or the persistent storage type preference into global KubeDirector settings -- maybe allowing override at app deployment time, but typically that would not be necessary. So when we do that we may simplify these example clusters so that we don't have a GKE-specific example and etc. (Although if we wanted to have some DOK quickstart documentation similar to our GKE doc, that would continue to be useful.)

joel-bluedata:
We're going to address this in a slightly different way as per PR #128.

Issue #85, opened 2018-10-10
differentiate nonspecified setup_package from emptystring


joel-bluedata:
The app CR author should be allowed to specify emptystring for a role's setup package in order to indicate that the role has no setup package, even when there is a top-level default package.

Currently though we can't differentiate emptystring from unspecified, in the field of the deserialized NodeRole structure. So we should change that field type from string to string-pointer and make the distinction.

For consistency we *could* change the top-level setup_package field, in AppSpec, to string-pointer type as well. The question really is in how we choose to treat the case where neither the AppSpec nor the NodeRole have setup_package specified. Do we consider that to be a mistake and flag it as an error? Or do we treat it as "no setup package", i.e. the same way as if one or both of those fields had been emptystring? if the latter, then we can leave the AppSpec field as string type.

joel-bluedata:
We also need to allow emptystring in the schema pattern, and document its behavior on the wiki.

(Unless there's a better way to indicate "no setup package"?)

joel-bluedata:
(@swamibluedata  reminded me that the setup_package value is actually an object rather than just the import_url string.)

At this point I'm thinking that we need to always make the "no setup package" case explicit, for clarity. I.e. at the top level you must always specify some value for setup_package, and at the role level you can omit setup_package to indicate "no override".

Some possible ways we could model the explicit "no setup package" case using the existing object:
* Have an alternate value for the setup_package property. For example allow it to be set to false rather than to the usual object value. However the current k8s validation schema isn't full-featured enough to support something like that (and it's a little odd). We could give up on schema validation here and just handle it dynamically...
* Or, have some other flag inside the object that explicitly indicates whether we are using a setup package or not. If true, you must specify both import_url and config_api_version. If false, specify neither. That's a little baroque.

As for the other approach...

We've been talking about dropping config_api_version since we don't need it yet. This is related to the question of how much we want the config API and app metadata stuff to track the implementation used for EPIC catalog entries... will we try to keep those more or less in sync as some global thing used across multiple products/projects? That would be nice in some ways but maybe too ambitious. We have a meeting scheduled to talk about some relevant topics so we might get more clarity on that soon. @kcsraju may want to chime in here.

Anyway, this is alpha code so we could drop the config_api_version complication now and bring it back later if we need it. I would be completely OK with that unless it is blindingly obvious that we will need it. IF we do that, then the approach should be:
- setup_package just takes a string value, which is the package location. Maybe we rename the property to setup_package_url.
[...]

Issue #77, opened 2018-10-08
member should not be notified if its role has no setup package


joel-bluedata:
We avoid running initial config or reconfig on a member if its role has no setup package. However, when we blast out add/del notifies we hit ALL ready members.

That's wrong. It's not fatal, since notifies are currently best effort, but wrong. We should skip members that are in roles that don't have setup packages.

Sort of related to issue #66... or at least, this type of thing would be clearer if we were handling the ready members in that proposed fashion.

joel-bluedata:
Can't fix/test this until the validation schema is adjusted to allow "no setup package" as mentioned in issue #85.

PR #118, opened 2018-12-08
Handle roles with no appconfig

kcsraju:
Resolves: https://github.com/bluek8s/kubedirector/issues/85 https://github.com/bluek8s/kubedirector/issues/77

### The possibilities
- Each role can specify it's own appconfig package. 
- If a specific role doesn't have any appconfig scripts `setup_package: null` must be used in the APP CR JSON. 
- If a role doesn't have any `setup_package` specification then the top level `setup_pacakge` specification is used by default. 
- Setting the toplevel `setup_package` to `null` is allowed. 
- If the `setup_package` is not specified anywhere in the APP CR JSON, then its treated as if no roles in the app have any setup packages. 

### Schema version
 `config_api_version` is now retired.  `schema_version` indicates both the JSON metadata version as well as the nodeprep/configcli API version.

### Sample: CentOS 7.x utility
I added a CentOS7x utility app with no setup_package as a sample for this case. The blurb below is from `kubectl describe kubedirectorcluster centos7-lb-persistent` after create, expand and shrink ops.

```
  Type    Reason   Age              From          Message
  ----    ------   ----             ----          -------
  Normal  Role     6m               kubedirector  creating role{utility}
  Normal  Role     6m               kubedirector  changing replicas count for role{utility}: 0 -> 1
  Normal  Cluster  6m               kubedirector  new
  Normal  Member   5m               kubedirector  initial config skipped for member{kd-cj8tw-0} in role{utility}
  Normal  Role     4m               kubedirector  expanding role{utility}
  Normal  Role     4m               kubedirector  changing replicas count for role{utility}: 1 -> 3
  Normal  Member   4m               kubedirector  initial config skipped for member{kd-cj8tw-1} in role{utility}
  Normal  Member   3m               kubedirector  initial config skipped for member{kd-cj8tw-2} in role{utility}
  Normal  Role     1m               kubedirector  shrinking role{utility}
  Normal  Role     1m               kubedirector  changing replicas count for role{utility}: 3 -> 2
  Normal  Member   1m (x3 over 4m)  kubedirector  notify skipped for member{kd-cj8tw-0} in role{utility}
  Normal  Member   1m (x2 over 3m)  kubedirector  notify skipped for member{kd-cj8tw-1} in role{utility}
  Normal  Cluster  1s (x3 over 5m)  kubedirector  stable
```

When the setup_package is explicitly set to `null` K8S show `<nil>` in the description

```
Name:         centos7x
Namespace:    default
Labels:       <none>
Annotations:  <none>
API Version:  kubedirector.bluedata.io/v1alpha1
Kind:         KubeDirectorApp
Metadata:
  Cluster Name:
  Creation Timestamp:  2018-12-08T22:10:54Z
  Resource Version:    121060
  Self Link:           /apis/kubedirector.bluedata.io/v1alpha1/namespaces/default/kubedirectorapps/centos7x
  UID:                 21986769-fb36-11e8-95b2-42010a8e0027
Spec:
  Config:
    Role _ Services:
      Role _ Id:  utility
      Service _ Ids:
        ssh
    Selected _ Roles:
      utility
  Distro _ Id:  bluedata/centos7x
  Label:
    Description:  CentOS 7x utility with no preinstalled apps
    Name:         CentOS 7.x utility.
  Roles:
    Cardinality:  1+
    Id:           utility
    Image:
      Repo Tag:      docker.io/bluedata/centos7:4.1
  Schema _ Version:  7
  Services:
    Endpoint:
      Is _ Dashboard:  false
      Port:            22
    Id:                ssh
    Label:
      Name:          SSH
  Setup _ Pacakge:   <nil>
  Systemd Required:  true
  Version:           1.0
Events:              <none>
```

### Embedded appconfig files
The existing `setup_package.import_url` can also be used to specify `file://` type urls so, I modified the regex to allow `file://` schema as well.

### Known issues
- Even if the role explicitly says it needs no setup package we still end up pushing the config meta.


deploy/example_clusters/cr-cluster-centos7-gke.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: "kubedirector.bluedata.io/v1alpha1"
+kind: "KubeDirectorCluster"
+metadata:
+  name: "centos7-lb-persistent"
+spec:
+  app: centos7x
+  serviceType: LoadBalancer
```
swamibluedata: we can remove this.

swamibluedata:
I think we should do this from our mutation webhook. That can also handle special case of being "null" string. So each role object's setup_url will be automatically populated (either set to empty string) or the actual value by our webhook (specified explicitly by the user or use app's default). Rest of the code base can just check for setup_url and decide whether to setup app config or not.

swamibluedata:
You can set ImportUrl as a string pointer and set it to nil from the webhook, if it decides that there is no setup package needed for the role

joel-bluedata:
The thing to consider about the mutation webhook is that it will actually change the submitted CR. So far, we've just been using that to populate static defaults for unspecified properties.

This case feels a little different.

For example, let's consider:
- I have an app CR that defines two roles A and B.
- The app CR says that the "global" app package URL is "foo".
- The app CR says that the app package URL for role A is "bar".
- The app package URL for role B has no specified app package; i.e. it should inherit the global package.

If we create this app CR, then edit the global app package to be "potato", what do we expect to be used as an app package for role B when deploying a cluster?

If we didn't mutate the object, then role B members will get app package "potato".

If we did mutate the object, then role B members will get app package "foo".

The correct approach depends on whether we think that the behavior of "use the global package" is an inherent and persistent characteristic of role B, or whether it was just a shortcut we used at the time of submitting the CR.

joel-bluedata:
Swami and I just talked briefly about this, and he did raise the good point that editing an app CR is not something that will happen during normal usage. It will happen during app CR development (probably) but it could be fine to have unique semantics for that.

Let me take a closer look to see how many code shenanigans would be obviated by using the mutator.

deploy/kubedirector/crd-app.yaml
```diff
@@ -31,21 +31,21 @@ spec:
             version:
               type: string
               minLength: 1
+            schema_version:
```
joel-bluedata: Can/should we give this a slightly more descriptive name? I.e. what is this the schema of (otherwise it seems like it's talking about this CR format).

pkg/apis/kubedirector.bluedata.io/v1alpha1/types.go
```diff
@@ -184,8 +188,8 @@ type ServiceEndpoint struct {
 type NodeRole struct {
 	ID           string       `json:"id"`
 	Cardinality  string       `json:"cardinality"`
-	Image        Image        `json:"image,omitempty"`
-	SetupPackage SetupPackage `json:"setup_package,omitempty"`
+	Image        Image        `json:"image_repo_tag,omitempty"`
+	SetupPackage SetupPackage `json:"config_package_url,omitempty"`
```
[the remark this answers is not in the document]
kcsraju: yeah! that makes sense.

pkg/apis/kubedirector.bluedata.io/v1alpha1/types.go
```diff
@@ -149,15 +150,23 @@ type Label struct {
...
 
 // SetupPackage describes the app setup package to be used. A top-level
 // package can be specified, and/or a role-specific package that will override
 // any top-level package.
 type SetupPackage struct {
-	ConfigAPIVersion int    `json:"config_api_version"`
-	ImportUrl        string `json:"import_url"`
+	IsSet      bool
+	IsNull     bool
+	PackageURL SetupPacakgeURL
```
swamibluedata: typo.

pkg/apis/kubedirector.bluedata.io/v1alpha1/types.go
```diff
@@ -149,15 +150,23 @@ type Label struct {
...
+}
+
+// SetupPacakgeURL is the URL of the setup package.
+type SetupPacakgeURL struct {
```
swamibluedata: ah... thats why it works.

pkg/validator/types.go
```diff
@@ -63,4 +63,6 @@ const (
 
 	invalidRoleStorageClass   = "Unable to fetch storageClassName(%s) for role(%s)."
 	undefinedRoleStorageClass = "storageClassName is not specified for one or more roles. Default storage class (%s) is not available on the system."
+
+	noImageForRole = "No container image is specified for role %s"
```
swamibluedata: there is already noDefaultImage. May be remove that?

pkg/reconciler/members.go
```diff
@@ -602,6 +595,30 @@ func notifyReadyNodes(
 			for _, member := range ready {
 				go func(m *kdv1.MemberStatus, r *roleInfo) {
 					defer wgReady.Done()
+
+					setupURL, setupURLErr := catalog.AppSetupPackageUrl(cr, r.roleStatus.Name)
+					if setupURLErr != nil {
```
swamibluedata: since this is a role based property, can we move into the outer role loop

swamibluedata: and also skip doing looping members if there is no setup url

pkg/catalog/interrogate.go
```diff
@@ -193,21 +191,23 @@ func AppSetupPackageUrl(
 		return "", err
 	}
 
-	var appConfigUrl string
+	var appConfigURL string
+	appConfigURL = ""
```
swamibluedata: nit:

var appConfigURL = ""

deploy/example_catalog/cr-app-cassandra-3.11.json
```diff
@@ -67,9 +66,8 @@
                 }
             }
         ],
-        "setup_package": {
-            "config_api_version": 7,
-            "import_url": "https://s3.amazonaws.com/bluek8s/kubedirector/cassandra311/appconfig-3.1.tgz"
+        "config_package_url": {
```
swamibluedata: should be config_package

deploy/example_catalog/cr-app-cdh5142cm.json
```diff
@@ -237,9 +236,8 @@
                 }
             }
         ],
-        "setup_package": {
-            "config_api_version": 7,
-            "import_url": "https://s3.amazonaws.com/bluek8s/kubedirector/cdh5142cm/cdh5-cm-setup-1.4.tgz"
+        "config_package_url":  {
```
swamibluedata: should be config_package

deploy/example_catalog/cr-app-spark221e2.json
```diff
@@ -104,29 +105,22 @@
                 }
             }
         ],
-        "setup_package": {
-            "config_api_version": 7,
-            "import_url": "https://s3.amazonaws.com/bluek8s/kubedirector/spark221e2/appconfig-2.6.tgz"
+        "config_package_url":  {
```
swamibluedata: should be config_package

swamibluedata: few more places too

deploy/example_catalog/cr-app-cassandra-3.11.json
kcsraju: I am changing all the other apps now and testing ... looks like I hit a roadblock. Gimme some time to get back to you.

kcsraju:
```
$ kubectl describe kubedirectorapp centos7x
...
  Config _ Package:           <nil>
  Config _ Schema _ Version:  7
  Distro _ Id:                bluedata/centos7x
  Image _ Repo _ Tag:         docker.io/bluedata/centos7:4.1
  Label:
    Description:  CentOS 7x utility with no preinstalled apps
    Name:         CentOS 7.x utility.
  Roles:
    Cardinality:         1+
    Config _ Package:    <nil>
    Id:                  utility
    Image _ Repo _ Tag:  docker.io/bluedata/centos7:4.1
    Cardinality:         0+
    Config _ Package:    <nil>
    Id:                  extra
    Image _ Repo _ Tag:  docker.io/bluedata/centos7:4.1
...
```

```
$> kubectl describe kubedirectorapp spark221e2
...
  Config _ Package:
    Package _ URL:            https://s3.amazonaws.com/bluek8s/kubedirector/spark221e2/appconfig-2.6.tgz
  Config _ Schema _ Version:  7
  Distro _ Id:                bluedata/spark221e2
  Label:
    Description:  Spark 2.2.1 with Jupyter
    Name:         Spark 2.2.1 on centos7x with Jupyter
  Roles:
    Cardinality:  1
    Config _ Package:
      Package _ URL:     https://s3.amazonaws.com/bluek8s/kubedirector/spark221e2/appconfig-2.6.tgz
    Id:                  controller
    Image _ Repo _ Tag:  docker.io/bluedata/sparkbase:2.0
    Cardinality:         0+
    Config _ Package:
      Package _ URL:     https://s3.amazonaws.com/bluek8s/kubedirector/spark221e2/appconfig-2.6.tgz
    Id:                  worker
    Image _ Repo _ Tag:  docker.io/bluedata/sparkbase:2.0
    Cardinality:         0+
    Config _ Package:
      Package _ URL:     https://s3.amazonaws.com/bluek8s/kubedirector/spark221e2/appconfig-2.6.tgz
    Id:                  jupyter
    Image _ Repo _ Tag:  docker.io/bluedata/jupyter:2.2
...
```

All existing apps have been tested as well.

pkg/reconciler/members.go
swamibluedata: We should still move this check to outer loop no?
Just use otherRole.roleStatus.Name to fetch and check if setup url is present and we can skip the whole iteration

kcsraju: The problem was doing so needs a change to the line `wgReady.Add(totalReady)`.

kcsraju: I moved the call to get the app setup package out of the member loop. At least that way, we only make the call once per role. 

I don't see a point in further optimizing this function unless I am wrong about the `go func` being inexpensive.

pkg/apis/kubedirector.bluedata.io/v1alpha1/decode.go
```diff
@@ -0,0 +1,81 @@
...
+
+package v1alpha1
+
+import "encoding/json"
+
+// UnmarshalJSON for SetupPackage handles the unmarshalling of three
+// scenarios wrt 'setup_package':
+//   1. omitted                 : IsSet==false
+//   2. explicitly set to null  : IsSet==true && IsNull==true
+//   3. Set to a valid object   : IsSet=true && IsNull==false
```
joel-bluedata: should be `IsSet==true` in the comment here (double equal)

pkg/apis/kubedirector.bluedata.io/v1alpha1/decode.go
```diff
@@ -0,0 +1,81 @@
...
+func (setupPackage *SetupPackage) UnmarshalJSON(
+	data []byte,
+) error {
+
+	// The fact that we entered this function means the filed is set otherwise,
```
joel-bluedata: typo `filed`

pkg/apis/kubedirector.bluedata.io/v1alpha1/decode.go
```diff
@@ -0,0 +1,81 @@
...
+// UnmarshalJSON for Image handles the unmarshalling of three
+// scenarios wrt 'image_repo_tag':
+//   1. omitted                 : IsSet==false
+//   2. explicitly set to null  : IsSet==true && IsNull==true
+//   3. Set to a valid object   : IsSet=true && IsNull==false
+func (image *Image) UnmarshalJSON(
+	data []byte,
+) error {
+
+	// The fact that we entered this function means the filed is set otherwise,
```
joel-bluedata: typo `filed`

pkg/apis/kubedirector.bluedata.io/v1alpha1/decode.go
```diff
@@ -0,0 +1,81 @@
...
+
+	if err := json.Unmarshal(data, &image.RepoTag); err != nil {
+		return err
+	}
+	image.IsNull = false
+
+	return nil
+}
+
+// // MarshalJSON is to handle encoding the Image into json.
```
joel-bluedata: can remove these comments?

pkg/apis/kubedirector.bluedata.io/v1alpha1/decode.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	setupPackage.IsNull = false
+
+	return nil
+}
+
+// UnmarshalJSON for Image handles the unmarshalling of three
+// scenarios wrt 'image_repo_tag':
+//   1. omitted                 : IsSet==false
+//   2. explicitly set to null  : IsSet==true && IsNull==true
```
joel-bluedata: We can't set the image to null, can we?

joel-bluedata: Going to pause reviewing until I understand this part.

kcsraju: No, the image cannot be set to null. We ensure an image is specified for every role during validation.  This part allows the global to be null (or unspecified) and each role to use its own spec.

joel-bluedata: The image property is a string value, so I don't think it can be set to null. If it is unspecified then that is handled separately (with IsSet==false) so I am still uncertain about why we are trying to support a null value here.

kcsraju: Yup! I see what you are saying. Fix is on the way!

pkg/catalog/interrogate.go
```diff
@@ -172,15 +172,13 @@ func ImageForRole(
 			break
```
joel-bluedata: At this point, we don't even need to declare repoTag outside the loop. We could just return immediately once we find the role, since the role-specific image will always be defined now. And the comment above the loop can be changed to something like "// Find the image for the role." or just deleted altogether.

Interestingly, if we do manage to get all the way through the loop without finding the role, something is wrong. In that case we should `return "", err`

pkg/catalog/interrogate.go
```diff
@@ -193,21 +191,22 @@ func AppSetupPackageUrl(
...
 		}
 	}
-	// If role-specific package is not present, use the main one.
-	if appConfigUrl == "" {
-		appConfigUrl = cr.AppSpec.Spec.SetupPackage.ImportUrl
-	}
 
-	return appConfigUrl, nil
+	return appConfigURL, nil
 }
```
joel-bluedata: And the situation with this loop is different since it's possible to find the role but still have no setup package. So this loop should be ok as is.

pkg/validator/app.go
```diff
@@ -122,26 +146,102 @@ func validateSelectedRoles(
...
+							packageURL: &packageURL{URL: *globalSetupPackageURL},
+						},
+					},
 				)
-				valErrors = append(valErrors, invalidMsg)
 			}
 		}
+
+		// We allow roles to have different container images but unlike the
+		// setup pacakge there cannot be a role with no image.
```
joel-bluedata: typo `pacakge`

joel-bluedata:
Done w/ review pass. One other question: if the package URL is a file URL, when we "curl"-fetch it does that "just work"?

pkg/catalog/interrogate.go
kcsraju: By this time, yes. I guess you are wondering if we can remove that check? Let me see.

kcsraju:
Yes, about curl and `file://` URL. It just works no other modifcation necessary. Although, it would be nice if we allow non-tarball files. But that is a different discussion. 

```
# curl file:///etc/hosts
# Kubernetes-managed hosts file.
127.0.0.1	localhost
::1	localhost ip6-localhost ip6-loopback
fe00::0	ip6-localnet
fe00::0	ip6-mcastprefix
fe00::1	ip6-allnodes
fe00::2	ip6-allrouters
10.32.4.9	kd-8r2xx-0.svc-xwpv9.default.svc.cluster.local	kd-8r2xx-0
```

pkg/reconciler/members.go
```diff
@@ -225,19 +225,27 @@ func handleCreatingMembers(
...
 	for _, member := range creating {
 		go func(m *kdv1.MemberStatus) {
 			defer wgSetup.Done()
-			// If setup package is not present, skip doing any setup.
+
 			if setupUrl == "" {
-				// Set a temporary state used below so we won't send notifies
-				// to this member yet.
-				m.State = string(memberConfigured)
+				m.State = string(memberReady)
```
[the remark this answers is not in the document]
kcsraju: The notification for roles with no config package will be done right away instead of waiting for another pass. That's the only reason.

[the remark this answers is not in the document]
joel-bluedata: This particular notification is to tell already-ready node A that new node B has been added. We don't want to notify node B about node B being added, so that's why we do the two-phase change to its readiness state.

In practice I think this still ends up working out OK since the notification code itself will skip notifying node B as it doesn't have a setup package. But anyway it seems like we should have just left the two-phase change as it was.

pkg/catalog/interrogate.go
kcsraju: Cleaned it up a bit.

joel-bluedata:
Woo!

PR #126, opened 2019-01-17
update deps

joel-bluedata:


joel-bluedata:
Just did a "make dep" and "make modules". Deployment, cluster creation, etc. still work.

PR #130, opened 2019-01-17
fix validation for storageclassname default

joel-bluedata:
If the value specified in the config was something other than "standard", this check was not working as desired. We should be checking whatever default is specified in the config here (rather than always checking "standard").


swamibluedata:
Ah... ok :shipit:

Issue #132, opened 2019-01-22
deal w/ persistent storage issues on EKS


joel-bluedata:
Probably need to set volumeBindingMode if/when EKS picks up support for K8s 1.12. It's unclear at the moment whether that will be entirely sufficient to make persistent storage work on EKS... maybe it will!

The issue: https://github.com/kubernetes/kubernetes/issues/34583

The solution: https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode

joel-bluedata:
Actually I misunderstood volumeBindingMode. That's a property on the StorageClass, not on the PVC. So if Amazon sets this, then things should "just work" for us when using 1.12 without any KD changes.

joel-bluedata:
Might get 1.12 on EKS soon: https://github.com/aws/containers-roadmap/issues/24

pawelprazak:
The current (`v1.12.6-eks-d69f1b`) default storage class does not configure this:
```
  apiVersion: storage.k8s.io/v1
  kind: StorageClass
  metadata:
    annotations:
      storageclass.kubernetes.io/is-default-class: "true"
    name: gp2
  parameters:
    fsType: ext4
    type: gp2
  provisioner: kubernetes.io/aws-ebs
  reclaimPolicy: Delete
  volumeBindingMode: Immediate
```

joel-bluedata:
Might be able to edit that resource to add that property now. That's something we should try.

joel-bluedata:
Assigning this to me to try it out.

joel-bluedata:
So, we can't edit that field on an existing storage class...it's immutable.

If I create a new storage class identical to the existing one, just with volumeBindingMode set to WaitForFirstConsumer (and with a new storage class name), that works fine. I can either set that as the default class with annotations, or I can set the KD config to explicitly use that class, and everything works.

If Amazon isn't going to set that by default then we need to figure out the best way to address it for KD users. Obviously we can/should document this issue -- maybe create a whole "eks-notes.md" in the docs -- but what else? Should KD reject attempts to use persistent storage if the designated class doesn't have this field setting (but then we have to detect whether we are on Amazon)?

joel-bluedata:
Plan is to document this with an eks-notes thing. We'll give explicit steps for setting up and using the new storage class.

PR #167, opened 2019-05-22
documentation and minor Makefile changes for EKS support

joel-bluedata:
addresses issue #132

### DIFF
--- a/deploy/example_config/eks-gp2-for-kd.yaml
+++ b/deploy/example_config/eks-gp2-for-kd.yaml
@@ -1,0 +1,31 @@
+# This is an example of one solution for the default StorageClass on EKS
+# not setting its volumeBindingMode to WaitForFirstConsumer. It creates
+# another storage class, which is identical to the "gp2" class except with
+# the necessary setting for volumeBindingMode, and then configures KD to use
+# that storage class (even though it is not the K8s default storage class).
+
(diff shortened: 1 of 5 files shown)


joel-bluedata:
Note that Travis CI builds are lagging behind because of a hiccup with GitHub webhooks. Hopefully ours will eventually kick in without further intervention.

deploy/example_config/eks-gp2-for-kd.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: gp2-for-kd
+  labels:
+    kubedirector-support: "true"
+parameters:
+  fsType: ext4
+  type: gp2
```
distortedsignal: It looks like both of these are the default value - do we need to include them? Can we leave the `parameters` block empty (or delete it entirely)?

joel-bluedata: Seems to still work when omitted, but no default values are filled in for the created resource, so it is a mystery what sort of PV it will create until it has been created. I think I'd rather leave it explicit.

distortedsignal: Fair.

Issue #38, opened 2018-10-02
prevent removal of our finalizer by non-KD clients?


joel-bluedata:
We could do some protection of changes to this property in a similar way as our protection of the status property.

That would be a nontrivial code addition. It's worth revisiting: what is the finalizer for, e.g. what bad things might happen if it were removed by someone else?

Currently we only make use of the finalizer to ensure that we don't try to update object status in a race with the object being deleted. We *could* make executor.UpdateStatus just be ok with not-found results, to obviate that issue... but it still could be a good thing in the future to have a finalizer that lets us do other post-cluster cleanup, so I'm reluctant to get rid of that finalizer just yet.

joel-bluedata:
FWIW, once PR #88 goes in it will no longer be a problem for the object to go away while we are trying to update its status.

joel-bluedata:
Not going to do this unless/until we get a good reason.

Issue #37, opened 2018-10-02
use sub-resource for status updates


joel-bluedata:
The status sub-resource requires a higher minimum K8s version that our current minimum, so we can't use it yet. Should we move to using it in the future? Not sure what the benefits would be of using that vs. our current patching.

joel-bluedata:
https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#status-subresource

It's available from 1.10, and enabled by default from 1.11. A CRD needs to explicitly be set up for using subresources in order for them to be enabled for that kind of CR.

One benefit seems to be that we could use RBAC to write-protect the status, instead of our current validator intervention.

joel-bluedata:
Bottom line, looks like a good thing, but we can't make this our method of doing status updates until our floor for Kubernetes version support is at least 1.10.

joel-bluedata:
Done.

PR #171, opened 2019-06-25
KubeDirector migrated to operator-sdk version v0.8.1

don-milos:
From the testing I've been able to do in my environment I believe the code is fully functional but there are still a number of issues that need to be cleaned up. I've marked all of the known issues with TODO: comments.

Per Joel's comments last week, I would like to get the majority of the changes checked in and clean the code up over time so that other people can work on this version of the code.


joel-bluedata:
I'm going to jump this PR to a non-master branch so when we eventually merge it, it can still have some ongoing development before we promote it to master.

build/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM alpine:3.6
```
joel-bluedata: I noticed that the current SDK generates a Dockerfile using ubi-minimal ([RedHat UBI](https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image)) by default. The original [motivation](https://github.com/operator-framework/operator-sdk/pull/1139) for this was apparently to make it easier to deploy on OpenShift -- although I'm not sure exactly why using UBI helps; maybe it's just a matter of support.

We should consider the pros and cons of moving to using UBI here. I think we will be using UBI for some of our other container deployments.

And more generally, here's what their current Dockerfile template looks like:
```
FROM registry.access.redhat.com/ubi7/ubi-minimal:latest
ENV OPERATOR=/usr/local/bin/{{.ProjectName}} \
    USER_UID=1001 \
    USER_NAME={{.ProjectName}}
COPY build/_output/bin/{{.ProjectName}} ${OPERATOR}
COPY build/bin /usr/local/bin
RUN  /usr/local/bin/user_setup
ENTRYPOINT ["/usr/local/bin/entrypoint"]
USER ${USER_UID}
```

Some of that (the user setup) seems moot as long as we continue to run as root. (Which we should try to get away from, and I have an idea about that, but that's a separate issue.) Other than that it APPEARS that the only interesting change is that base image. Concur?

joel-bluedata:
Do you remember where build.sh and docker_build.sh are invoked from? I noticed that the new SDK normally puts those files in build/bin/ but you have placed them just in build/.

Makefile
```diff
@@ -24,37 +26,58 @@ sedseparator = ''
...
-        -e 's~$$~              valueFrom:~; G; ' \
-        -e 's~$$~                fieldRef:~; G; ' \
-        -e 's~$$~                  fieldPath: metadata.namespace~;' \
-        -e '}' deploy/operator.yaml
+        -e 'a\            - name: MY_NAMESPACE' \
+        -e 'a\              valueFrom:' \
+        -e 'a\                fieldRef:' \
+        -e 'a\                  fieldPath: metadata.namespace' \
+        -e '}' \
+        -e 's~REPLACE_IMAGE~${image}~' \
```
joel-bluedata: Now that operator.yaml is checked into our repo rather than being generated, it seems like we can avoid most of this sed stuff (by having the desired text just already be in operator.yaml). Maybe the only thing we still need to change in operator.yaml at buildtime is REPLACE_IMAGE.

Makefile
```diff
@@ -24,37 +26,58 @@ sedseparator = ''
...
+        -e 'a\          - containerPort: 40000' \
+        -e 'a\            hostPort: 40000' \
+        -e '}' \
+        -e '$$ {' \
+        -e 'a\          securityContext:' \
+        -e 'a\            capabilities:' \
+        -e 'a\              add:' \
+        -e 'a\                - SYS_PTRACE' \
+        -e '}' \
+        ${local_deploy_yaml}
```
joel-bluedata: Nice. You may need to give us a talk about your debugging process.

Makefile
```diff
@@ -115,13 +141,13 @@ deploy:
         echo; \
         echo -n \* Waiting for KubeDirector to start...; \
         sleep 3; \
-        retries=20; \
-        while [ $$retries ]; do \
```
joel-bluedata: This change is fine, but curious about motivation. Not working correctly on your system?

joel-bluedata:
While we're messing around in the Makefile we should also fix the behavior of the "compile" target. It works ok as a sanity check done during CI, but it is broken as a target that you might want to invoke manually (doesn't cross-compile and writes to wrong file). It should do the same go build command as is done inside build.sh. Or maybe it should just invoke build.sh?

cmd/manager/main.go
```diff
@@ -0,0 +1,187 @@
...
+
+func printVersion() {
+	log.Info(fmt.Sprintf("Go Version: %s", runtime.Version()))
+	log.Info(fmt.Sprintf("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH))
+	log.Info(fmt.Sprintf("Version of operator-sdk: %v", sdkVersion.Version))
+}
+
+func main() {
+	// TODO: the operator-sdk switched to a new log implementation.
+	//       should we migrate the shared logger code?
```
joel-bluedata: Yeah. We don't have any particular attachment to logrus, and the fewer random thirdparty dependencies we have the better.

cmd/manager/main.go
```diff
@@ -0,0 +1,187 @@
...
+		os.Exit(1)
+	}
+
+	// Setup all Controllers
+	if err := controller.AddToManager(mgr); err != nil {
+		log.Error(err, "")
+		os.Exit(1)
+	}
+
+	// Create Service object to expose the metrics port.
```
joel-bluedata: Does this create a Service object that the Makefile teardown target should clean up? Or is the Service going to be owner-ref'd to the operator so that it gets automatically deleted?

joel-bluedata:
Do we really need the empty example CR yaml files (like deploy/kubedirector/kubedirector_v1alpha1_kubedirectorapp_cr.yaml) in the deploy/kubedirector directory?

deploy/kubedirector/kubedirector_v1alpha1_kubedirectorcluster_crd.yaml
```diff
@@ -75,7 +84,7 @@ spec:
                         value:
                           type: string
                   storage:
-                    required: [size]
+                    # required: [size]
```
joel-bluedata: Curious about this change. In cases where the storage property is populated with some object value, the size property in that object is indeed required.

deploy/kubedirector/kubedirector_v1alpha1_kubedirectorcluster_crd.yaml
```diff
@@ -10,10 +10,19 @@ spec:
     plural: kubedirectorclusters
     singular: kubedirectorcluster
   scope: Namespaced
-  version: v1alpha1
+  subresources:
+    status: {}
```
joel-bluedata: I'm not certain this is correct. We haven't been using status as an explicit "subresource", but rather just as a property of the object. Initially this was because we were running on at least some K8s versions that didn't support custom CR subresources. Cf. https://github.com/bluek8s/kubedirector/issues/37

We might be able to move to using a subresource now, but I'm not sure what code changes might then be required.

deploy/role.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
joel-bluedata: I don't think this role.yaml file is being used; the Makefile creates deploy/kubedirector/rbac.yaml from deploy/kubedirector/rbac-default.yaml and then uses that rbac.yaml file.

deploy/service_account.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: ServiceAccount
```
joel-bluedata: Similarly I don't think this file is being used either.

version/version.go
```diff
@@ -0,0 +1,5 @@
+package version
+
+var (
+	Version = "0.0.1"
```
joel-bluedata: I think this is intended to be our own product version. And maybe it should be printed in the printVersion function in main.go.

So we should change this to be 0.3.0 for now. And update the kd-release-process.md document to include the necessity of updating this file.

We could think about ways to get cleverer with this but I think that's fine for now.

tools.go
```diff
@@ -0,0 +1,15 @@
+// +build tools
+
+package tools
```
joel-bluedata: Hmm what is this file for?

pkg/controller/kubedirectorconfig/kubedirectorconfig_controller.go
```diff
@@ -0,0 +1,106 @@
...
+// blank assignment to verify that ReconcileKubeDirectorConfig implements reconcile.Reconciler
+var _ reconcile.Reconciler = &ReconcileKubeDirectorConfig{}
+
+// ReconcileKubeDirectorConfig reconciles a KubeDirectorConfig object
+type ReconcileKubeDirectorConfig struct {
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the apiserver
+	Client client.Client
+	scheme *runtime.Scheme
+	lock         sync.RWMutex
```
joel-bluedata: So originally we had one lock covering both clusterState and globalConfig. In this change, each of them has their own lock.

I think this is fine. Our only real consistency concerns were in making sure that an update to one of those items:
1) didn't allow a read of that same item until the update was done, and
2) didn't blow away an update to the other item that had just happened

As far as I know/remember there was no fundamental need for synchronizing access to both of them at the same time, from a semantic consistency standpoint. So I think this change is fine. Just talking it out here.

(I know you have other concerns about the right way to share access to the clusterState.)

pkg/controller/kubedirectorconfig/doc.go
```diff
@@ -17,4 +17,5 @@
 // The handler logic will make use of the other packages (observer, catalog,
 // executor, etc.) to determine the current state of the relevant resources
 // and adjust them to match the provided spec.
-package reconciler
+
+package kubedirectorconfig
```
joel-bluedata: Need to change the package name in the comments above ("Package reconciler").

cmd/manager/main.go
```diff
@@ -0,0 +1,187 @@
...
+	k8sCoreClient, _ := client.New(mgr.GetConfig(), client.Options{})
+
+	kd, err := observer.GetDeployment(kdName, k8sCoreClient)
+	if err != nil {
+		log.Error(err, "failed to get kubedirector deployment object")
+		os.Exit(1)
+	}
+
+	// TODO: migrate the current webhook code to the controllers and using the
+	//       new framework; i.e. mgr.Add(hookServer) and hookServer.Register()
```
joel-bluedata: I think we can fairly create an Issue for this, and delete the TODO comment.

pkg/controller/kubedirectorcluster/util.go
```diff
@@ -92,52 +95,36 @@ func ClustersUsingApp(
...
 ) bool {
-	handler.lock.RLock()
-	defer handler.lock.RUnlock()
-	if handler.globalConfig != nil && handler.globalConfig.Spec.NativeSystemdSupport != nil {
-		return *(handler.globalConfig.Spec.NativeSystemdSupport)
+	// Fetch global config CR (if present)
+	kdConfigCR, _ := observer.GetKDConfig(shared.KubeDirectorGlobalConfig, reconciler.Client)
+
+	if kdConfigCR != nil && kdConfigCR.Spec.NativeSystemdSupport != nil {
+		return *kdConfigCR.Spec.NativeSystemdSupport
```
joel-bluedata: Eh. Hmm...

The only reason we watch the config object and cache it is so that we can then use the cached value, which this function now circumvents.

Looking in our codebase at the other function that uses the globalConfig -- the cluster validator -- I see that it currently does the same thing, always fetching the config from K8s and ignoring the cache. So that's not ideal either.

Anyway, we have the same sort of issue here that we do with clusterState, i.e. how to best share some data structures among the different kinds of handler processes. We could punt on that in the case of globalConfig and just always fetch it (and rip out its CR reconciliation handler entirely), but I don't think that will get rid of the general problem... in fact we will probably want to add even more cached stuff in the form of the app catalog, so that the cluster reconciler doesn't have to fetch the app definition CR each time it runs.

I think it's fine to have one or more global data structures that contain this shared data and the necessary locking facilities. We can look into whether we want to have those directly accessed wherever they are needed, or else accessed at just the very top-level entrypoints of the processes and then the necessary data structures passed down the call chain as arguments.

pkg/controller/kubedirectorcluster/kubedirectorcluster_controller.go
```diff
@@ -0,0 +1,126 @@
...
+var log = logf.Log.WithName("controller_kubedirectorcluster")
+
+// Add creates a new KubeDirectorCluster Controller and adds it to the Manager.
+// The Manager will set fields on the Controller and Start it when the Manager is Started.
+func Add(mgr manager.Manager) error {
+	return add(mgr, newReconciler(mgr))
+}
+
+// TODO: this is a hack to make the cluster state info available
+//  to the validator. We need to figure out a better way.
```
joel-bluedata: Made some comments in pkg/controller/kubedirectorcluster/util.go about this. We can discuss further offline.

pkg/executor/service.go
```diff
@@ -93,17 +94,18 @@ func CreatePodService(
...
+	portInfoList, portsErr := catalog.PortsForRole(cr, role.Name, client)
 	if portsErr != nil {
 		return nil, portsErr
 	}
-	if len(portInfoList) == 0 {
-		return nil, nil
-	}
+    if len(portInfoList) == 0 {
+        return nil, nil
+    }
```
joel-bluedata: Indentation. Running "make format" should catch this kind of thing.

joel-bluedata:
Not done w/ review pass yet... will return for more later.

cmd/manager/main.go
```diff
@@ -0,0 +1,187 @@
...
+var (
+	metricsHost       = "0.0.0.0"
+	metricsPort int32 = 8383
+)
+var log = logf.Log.WithName("cmd")
+
+func printVersion() {
+	log.Info(fmt.Sprintf("Go Version: %s", runtime.Version()))
+	log.Info(fmt.Sprintf("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH))
+	log.Info(fmt.Sprintf("Version of operator-sdk: %v", sdkVersion.Version))
```
joel-bluedata: Could log our own version here too, as mentioned in comment for version.go.

cmd/manager/main.go
```diff
@@ -0,0 +1,187 @@
...
+	cfg, err := config.GetConfig()
+	if err != nil {
+		log.Error(err, "")
+		os.Exit(1)
+	}
+
+	ctx := context.TODO()
+
+	// Become the leader before proceeding
+	err = leader.Become(ctx, "kubedirector-lock")
```
joel-bluedata: Interesting. Seems like a good idea? Only protects from multiple KD deployments within a given namespace tho.

cmd/manager/main.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		log.Error(err, "")
+		os.Exit(1)
+	}
+
+	// Create a new Cmd to provide shared dependencies and start components
+	mgr, err := manager.New(cfg, manager.Options{
+		Namespace:          namespace,
```
joel-bluedata: So in current KD, when we watch for KubeDirectorCluster we set namespace to emptystring so that we end up watching all namespaces. For KubeDirectorConfig on the other hand we only watch in a single namespace. it's this code in the old main.go:
```go
	watchParams := []watchInfo{
		{
			kind: "KubeDirectorCluster",
			// The resync period essentially determines how granularly we can detect
			// the completion of cluster config changes. Making this too small can
			// actually be bad in that there is benefit to batch-resolving changes,
			// within KubeDirector but also especially with the cluster's app config
			// scripts.
			resyncPeriod: time.Duration(30) * time.Second,
			namespace:    "",
		},
		{
			kind:         "KubeDirectorConfig",
			resyncPeriod: 0,
			namespace:    namespace,
		},
	}
```

I'm not 100% sure about the behavior of this new setup -- what the exact implications are of passing this namespace to the manager -- but in any case we want to preserve the old behavior where we can watch for KubeDirectorCluster objects appearing in any namespace.

cmd/manager/main.go
```diff
@@ -0,0 +1,187 @@
...
+		os.Exit(1)
+	}
+
+	// TODO: determine how we can use mgr.GetClient. Currently the
+	//       manager's client doesn't recognize core objects.
+	k8sCoreClient, _ := client.New(mgr.GetConfig(), client.Options{})
+
+	kd, err := observer.GetDeployment(kdName, k8sCoreClient)
+	if err != nil {
+		log.Error(err, "failed to get kubedirector deployment object")
```
joel-bluedata: Wrong error message here.

pkg/executor/statefulset.go
```diff
@@ -63,21 +56,27 @@ func UpdateStatefulSetReplicas(
...
-		{
-			Op:    "replace",
-			Path:  "/spec/replicas",
-			Value: replicas,
-		},
-	}
-	replicasPatchBytes, patchErr := json.Marshal(replicasPatch)
-	if patchErr == nil {
-		patchErr = sdk.Patch(statefulSet, types.JSONPatchType, replicasPatchBytes)
+	// TODO: can we just use the statefulset that was passed in?
```
joel-bluedata: I think so. It either comes from observer.GetStatefulSet in initRoleInfo, or from executor.CreateStatefulSet in handleRoleCreate. Probably shouldn't re-fetch it here, to avoid consistency issues. Hopefully the new form of CreateStatefulSet is still returning a complete valid statefulset object.

And of course we really should move back to using a PATCH whenever we can move to a stable version of controller-runtime that supports that (v0.2.0 or later). We can note that in a new Issue.

pkg/executor/service.go
```diff
@@ -160,33 +163,48 @@ func UpdatePodService(
...
-		Path  string         `json:"path"`
-		Value v1.ServiceType `json:"value"`
-	}
 
-	patch := []patchSpec{
-		{
-			Op:    "replace",
-			Path:  "/spec/type",
-			Value: reqServiceType,
+	// TODO: should we just use the service that was passed in?
```
joel-bluedata: Similar comments here as for the other patch situation. In this case the service pointer originally comes from queryService (which, in the non-nil case, is returning the result of observer.GetService).

pkg/controller/kubedirectorconfig/kubedirectorconfig_controller.go
```diff
@@ -0,0 +1,106 @@
...
+func (r *ReconcileKubeDirectorConfig) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling KubeDirectorConfig")
+
+	// Fetch the KubeDirectorConfig instance
+	kdConfig := &v1alpha1.KubeDirectorConfig{}
+	err := r.Client.Get(context.TODO(), request.NamespacedName, kdConfig)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
```
joel-bluedata: So just to confirm... are we always going to get a reconcile request when deletion happens? I.e. this case is expected when deletion happens (and is not just from a modify/deletion race).

joel-bluedata:
I have to pause my review again for a bit. I really only have these three files to look at, although I've been intentionally saving the more interesting ones to the last:
* pkg/controller/kubedirectorcluster/cluster.go
* pkg/controller/kubedirectorcluster/kubedirectorcluster_controller.go
* pkg/executor/cluster.go

joel-bluedata:
There's stuff to discuss around "status gen" UUIDs and valid-markers. I'm going to type up a description of this functionality here for both of our benefits, because it's been a while since I set up this arrangement and I need to wake up those brain cells.

The status gen tracking served two purposes:

1) We don't want clients other than KubeDirector making changes to the status block of a KDC resource.

2) We don't want to process a stale poll update from the framework.

Both of those purposes require a bit more discussion.

----------

For item 1:

The status block of a KDC resource is essentially KD's persistent state storage for managing that resource. If that got corrupted then things would break. So we don't want other clients writing to the status block. But clients should be allowed to change the spec and metadata.

If the status was a "subresource" then it could be protected with ACLs, but a) that's kind of fragile as it relies on the K8s system being configured correctly and staying correct every time ACLs are created or modified, and b) originally we supported running on versions of K8s that didn't support subresources for CRs. Issue "b" there is now a non-issue, but issue "a" is still worth thinking about.

The general implementation of this is that the the reconciler and validator share access to the "status gen" data for that KDC in memory. When the reconciler wants to make an update to the status, it can store in memory the expected value of a special GenerationUID property that it is going to include in its write. When the validator sees an incoming write that changes the status, it can check to see whether the GenerationUID value is as expected. If not then the write will be rejected.

The "validated" flag is used to make sure that we only accept one change for any given GenerationUID.

This all happens in validator.admitClusterCR in the old code.

I don't think we want to move to using subresources quite yet, both because of issue "a" above and also because that would probably involve a bit more code churn and unknown knock-on issues on top of all this other churn. So I think the status gen is still important to have.

(If it does turn out to be problematic to keep doing status-protection this way however, then we may need to look at moving over to the use of subresources.)

----------

For item 2:

The operator SDK's framework fires off the reconciler handler when a watched resource changes. It also just fires the handler periodically. It's possible in this case to have one of those periodic "refresh" polls coming in at about the same time that a previously triggered handler is writing some changes to the resource.

With the old framework, the resource itself would be passed in along with the handler invocation. So in the case described above, the incoming poll triggering the handler could be carrying a version of the resource that did not include the changes that we just wrote into the resource. It is stale and should be ignored.

We can check the GenerationUID value to see if it is the most recent value that we remember writing to the resource. If it isn't, then this is a stale polling-driven update and we should bail out of the handler and do nothing. This happens in reconciler.handleStatusGen in the old code.

With the new framework, the resource is not given to us and we have to fetch it ourselves. This removes the possibility of this stale-resource race and it looks to me like item 2 no longer matters.

That does NOT mean that handleStatusGen can just be thrown away, as it does a few different tasks.


----------

Given that item 2 is out the window, we could rename handleStatusGen to something like handleNewCluster and simplify it down to this (untested example code):
```go
// handleNewCluster looks in the cache for the last-known status generation
// UID for this CR. If there is one, return true to keep processing the CR.
// If there is not any last-known UID, this is either a new CR or one that
// was created before this KD came up. In the former case, where the CR status
// itself has no generation UID: set the cluster state to creating (this will
// also trigger population of the generation UID) and return false to cause
// this handler to exit; we'll pick up further processing in the next handler.
// In the latter case, sync up our internal state with the visible state of
// the CR and return true to continue processing.
func handleNewCluster(
    cr *kdv1.KubeDirectorCluster,
    reconciler *ReconcileKubeDirectorCluster,
) bool {

    lastKnown, ok := ReadStatusGen(cr, reconciler)
    if ok {
        return true
    }
    incoming := cr.Status.GenerationUID
    if incoming == "" {
        shared.LogInfo(
            cr,
            shared.EventReasonCluster,
            "new",
        )
        cr.Status.State = string(clusterCreating)
        return false
    }
    shared.LogWarnf(
        cr,
        shared.EventReasonNoEvent,
        "unknown with incoming gen uid %s",
        incoming,
    )
    writeStatusGen(cr, reconciler, incoming)
    ValidateStatusGen(cr, reconciler)
    ensureClusterAppReference(cr, reconciler)
    return true
}
```

----------

Now let's turn to the issue you noted in a comment in kubedirectorcluster.Reconcile. It's no longer possible to do deleteStatusGen when the resource has been deleted, because in the new framework we don't know the resource UID of the deleted CR.

One possibility is to index the cache by a resource's NamespacedName rather than by UID. This would work fine in 99.999% of the situations, but there's a tiny race window where another CR with the same NamespacedName could be created and processed before we process the delete of the previous resource.

We can instead do deleteStatusGen in handleFinalizers when we are removing our finalizer. Once a CR has been deletion-timestamped we will no longer do anything interesting with its status block, and so the status gen is no longer important. Especially since we don't have to worry about stale poll updates anymore.

This does raise the question: what happens if someone else removes our finalizer? That's an issue in any situation where a finalizer is actually useful, so I guess it's just something you live with. Should the KDC validator explicitly prevent other clients from removing the KD finalizer (similar to how we protect the status block)? I thought about that a little in issue #38 and decided not to implement that protection yet, but maybe that's a decision we should revisit.

Anyway, let's say we move deleteStatusGen into handleFinalizers, right after the call to RemoveFinalizer. Like this (untested example code):

```go
// handleFinalizers will remove our finalizer and clear the cached status gen
// if deletion has been requested. Otherwise it will add our finalizer if it
// is absent.
func handleFinalizers(
    cr *kdv1.KubeDirectorCluster,
    reconciler *ReconcileKubeDirectorCluster,
) (bool, error) {

    if cr.DeletionTimestamp != nil {
        // If a deletion has been requested, while ours (or other) finalizers
        // existed on the CR, go ahead and remove our finalizer.
        removeErr := executor.RemoveFinalizer(cr, reconciler.Client)
        if removeErr == nil {
            shared.LogInfo(
                cr,
                shared.EventReasonCluster,
                "greenlighting for deletion",
            )
        }
        // Also clear the status gen from our cache, regardless of whether
        // finalizer modification succeeded.
        deleteStatusGen(cr, reconciler)
        return true, removeErr
    }

    // If our finalizer doesn't exist on the CR, put it in there.
    ensureErr := executor.EnsureFinalizer(cr, reconciler.Client)
    if ensureErr != nil {
        return true, ensureErr
    }

    return false, nil
}
```

----------

Last thing: With that change we now have the situation where we could get a poll-triggered handler for a KDC CR that has a deletion timestamp and no finalizer, which has no entry in our status gen cache. We need to make sure that is handled correctly, i.e. ignored.

So in the KDC reconciler, we want to make sure we don't do any processing for a CR that has a deletion timestamp, beyond (if necessary) removing its finalizer and removing its status gen from the cache. It looks like handleFinalizers already takes care of that. However, the current ordering of the calls to handleNewCluster (used to be handleStatusGen) and handleFinalizers is unfortunate in that handleNewCluster would re-add the status gen back into the cache. I *believe* the simplest answer to that problem is to switch the order that those functions are called -- always do handleFinalizers first. I think the reason that the invocations were in their current order was to properly ignore stale poll handlers (which is no longer an issue).

I'm wondering if we need any change to the KDC validator. I think the validator will be fine as-is, since we won't be trying to make any changes to the status block after clearing the cached status gen UID. In fact the deferred func in syncCluster that does status update will explicitly do nothing if the deletion timestamp is set. But if for some reason the validator gets cantankerous about something related to these changes, we can look into fixes that change its behavior for CRs that have a deletion timestamp.

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -103,14 +91,14 @@ func syncCluster(
 		}
 	}()
 
-	// Ignore stale poll-driven handler for a resource we have since
+	// Ignore stale poll-driven reconciler for a resource we have since
```
joel-bluedata: As per my super-long review comment, this code comment will need to be changed a bit. And the ordering of this call with handleFinalizers below should probably be switched.

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -103,14 +91,14 @@ func syncCluster(
...
 	// updated. Also for a new CR just update the status state/gen.
-	shouldProcessCR := handleStatusGen(cr, handler)
+	shouldProcessCR := handleStatusGen(cr, reconciler)
 
 	// Regardless of whether the status gen is as expected, make sure the CR
```
joel-bluedata: This code comment will need to be tweaked too.

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -199,11 +188,11 @@ func syncCluster(
 // reject old/stale versions of the CR.
 func handleStatusGen(
```
joel-bluedata: Some changes to this function proposed in my long comment.

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -242,12 +231,13 @@ func handleStatusGen(
 // Otherwise it will add our finalizer if it is absent.
 func handleFinalizers(
```
joel-bluedata: Some changes to this function proposed in my long comment.

pkg/controller/kubedirectorcluster/kubedirectorcluster_controller.go
```diff
@@ -0,0 +1,126 @@
...
+		// the finalizers have been removed, and the kubedirectorcluster
+		// resource has been deleted, so there is nothing left
+		// to do.
+		if apierrors.IsNotFound(err) {
+			// TODO: I put this code here temporarily to make the behaviour
+			//  of the existing code, but we don't have kdCluster's UID at
+			//  this point because the cluster has been deleted.
+			//  deleteStatusGen() needs to be called or we will leak entries
+			//  in the map.
+			// deleteStatusGen(kdCluster, r)
```
joel-bluedata: Addressed in my long comment.

I also wondered whether we should move the invocation of removeClusterAppReference. Not sure? It's kind of nice that in this case we can do this cleanup when the object is for-sure-for-sure deleted, but it might be best to keep the cleanup activities bundled together.

If we did move it to handleFinalizers, then we would also have to modify syncCluster a bit so that it doesn't do ensureClusterAppReference until AFTER any early exit triggered by handleFinalizers.

joel-bluedata:
Some discussion of the problems with controller-runtime not yet supporting PATCH:

The potential gotcha in all of these cases (where we try to do a PATCH without actually doing a PATCH) is that there could be a race where some other client has updated the object since we last fetched it, and if we write back an object that we've been holding in memory then our write can fail -- and if we take steps to make it not fail (more on that in a minute) we can be clobbering someone else's changes.

Doing a GET of the object *right before* we change and update the object does reduce that race window. Although it doesn't technically eliminate it.

For objects like services that KD creates, I'm fine with clobbering someone else's changes, because other clients should not be modifying the objects anyway. For a KDC however we do want and expect clients to change the object.

So we should have awareness that the race can happen, and we really should move to using PATCH when controller-runtime enables that. Or move to using a status subresource, which has been discussed elsewhere -- although that would only help us for status-block updates and doesn't help us when it comes to doing finalizer updates and other stuff.

----------

What does that mean we should do for all these wannabe-PATCH cases?

I think we should get some shared clarity on what is up with controller-runtime. The 0.2.0 milestone on their GitHub only has one remaining issue, but I don't know if that's necessarily an indication of an imminent stable release. Possibly we could/should use a beta release? But there might also be unknown issues with moving our whole build (including the framework) over to that new version.

Unless it looks really tempting to switch over to 0.2.0 ASAP, we should implement a common PATCH utility function in our code. It could take as input a) the current known object and b) a func for appropriately modifying the object to apply the change.

Now as mentioned above, even if we do a GET right before the PUT, there's a race possibility. And I think the CR that we receive from a GET will include a metadata.resourceVersion value, which automatically changes on every write. If someone else sneaks in an update to the CR before we do our PUT, our PUT will be rejected because it has a non-matching resourceVersion.

Maybe we should do something like:
* Rather than starting out with a GET, just change the initial input object and try to PUT that.
* If that is rejected because of non-matching resourceVersion, do a GET to fetch the more recent object, and try changing and PUT-ting from that.
* If THAT also fails because of non-matching resourceVersion, hmm. We'd need to discuss what the retry and final error behavior is.

pkg/executor/statefulset.go
joel-bluedata: More discussion of PATCH stuff in my long comment above.

pkg/executor/service.go
joel-bluedata: More discussion of PATCH stuff in my long comment below.

pkg/executor/cluster.go
```diff
@@ -15,79 +15,68 @@
...
-	patchErr := patchAttempt(statusPatch)
-	if patchErr != nil {
-		// If replace doesn't work, try add. (First time update.)
-		statusPatch[0].Op = "add"
-		patchErr = patchAttempt(statusPatch)
-	}
-	if patchErr != nil {
+
+	// TODO: Can we just use the cr that was passed in?
+	prevCr.Status = cr.Status
```
joel-bluedata: More comments about PATCH stuff above.

pkg/executor/cluster.go
```diff
@@ -102,32 +91,44 @@ func RemoveFinalizer(
 		return nil
 	}
 
-	finalizersPatch := []finalizersPatchSpec{
-		{
-			Op:    "replace",
-			Path:  "/metadata/finalizers",
-			Value: cr.Finalizers,
+	// TODO: Can we just use the cr that was passed in?
+	prevCr := &kdv1.KubeDirectorCluster{}
```
joel-bluedata: More comments about PATCH stuff above.

pkg/executor/cluster.go
```diff
@@ -140,44 +141,44 @@ func EnsureFinalizer(
...
-				Path:  "/metadata/finalizers",
-				Value: cr.Finalizers,
-			},
-		}
-	}
-	finalizersPatchBytes, patchErr := json.Marshal(finalizersPatch)
-	if patchErr == nil {
-		patchErr = sdk.Patch(cr, types.JSONPatchType, finalizersPatchBytes)
+	// TODO: Can we just use the cr that was passed in?
+	prevCr := &kdv1.KubeDirectorCluster{}
```
joel-bluedata: More comments about PATCH stuff above.

joel-bluedata:
Done w/ review pass. I'm also going to ask Swami to take a look in here, at least at the particular spots I commented on.

Once we all chew on these issues a bit and knock out any low-hanging fruit and major changes that we all agree on... we can talk about which remaining things really need to be tackled before at least settling this into its branch here, and which things can just be recorded as new issues. Then I can look at getting this branch up to sync with the current master.

joel-bluedata:
One more thing. In the old framework we could specify the resyncPeriod for watching a particular CR, so that we could control how frequently our handler might get triggered. Is there an equivalent in the new framework?

build/Dockerfile
swamibluedata: Will we get into subscription issues, if we have to do install any other software?

Makefile
```diff
@@ -24,37 +26,58 @@ sedseparator = ''
...
+debug: export GOFLAGS += -gcflags=all=-N -gcflags=-l
+debug: build
+debug:
+	@echo \* Applying debug settings to KubeDirector deployment YAML...
+	@sed -i ${sedseparator} \
+        -e '/\/root\/${project_name}/s~~/root/dlv --listen=:40000 --headless --api-version=2 exec &~' \
+        -e '/template:/,/metadata:/ {' \
+        -e '/metadata:/ {' \
+        -e 'a\      annotations:' \
+        -e 'a\        container.apparmor.security.beta.kubernetes.io/${project_name}: unconfined' \
```
swamibluedata: for selinux enabled system, we may have to do selinux disabled as well

The equivalent docker run is this
--security-opt label=disable

But need to check on k8s

Makefile
```diff
@@ -24,37 +26,58 @@ sedseparator = ''
...
+        -e '}' \
+        -e '}' \
+        -e '/image:/ {' \
+        -e 'a\          ports:' \
+        -e 'a\          - containerPort: 40000' \
+        -e 'a\            hostPort: 40000' \
```
swamibluedata: nit: What would it take to make this NodePort instead of hostPort

cmd/manager/main.go
```diff
@@ -0,0 +1,187 @@
...
+
+	// TODO: migrate the current webhook code to the controllers and using the
+	//       new framework; i.e. mgr.Add(hookServer) and hookServer.Register()
+	err = validator.InitValidationServer(
+		*metav1.NewControllerRef(kd,
+			// kd.GroupVersionKind()),
+			schema.GroupVersionKind{
+				Group:   kdv1.SchemeGroupVersion.Group,
+				Version: kdv1.SchemeGroupVersion.Version,
+				Kind:    "KubeDirectorCluster",
```
swamibluedata: This should be a group kind. Not just KubeDirectorCluster?

deploy/kubedirector/kubedirector_v1alpha1_kubedirectorcluster_crd.yaml
swamibluedata: yes size is required

deploy/operator.yaml
```diff
@@ -0,0 +1,33 @@
...
+      labels:
+        name: kubedirector
+    spec:
+      serviceAccountName: kubedirector
+      containers:
+        - name: kubedirector
+          # Replace this with the built image name
+          image: REPLACE_IMAGE
+          command:
+          - kubedirector
```
swamibluedata: Change to use the following
```          command:
          - "/bin/sh"
          args:
          - "-c"
          - "mkfifo /tmp/fifo; (/root/kubedirector &> /tmp/fifo) & while true; do cat /tmp/fifo; done"
```

pkg/apis/kubedirector.bluedata.io/v1alpha1/kubedirectorapp_types.go
```diff
@@ -0,0 +1,128 @@
...
+// NodeRole describes a subset of virtual cluster members that will provide
+// the same services. At deployment time all role members will receive
+// identical resource assignments.
+type NodeRole struct {
+	ID           string       `json:"id"`
+	Cardinality  string       `json:"cardinality"`
+	ImageRepoTag *string      `json:"image_repo_tag,omitempty"`
+	SetupPackage SetupPackage `json:"config_package,omitempty"`
+	PersistDirs  *[]string    `json:"persist_dirs"`
+	MinResources *v1.ResourceList `json:"min_resources"`
```
swamibluedata: should run make format

pkg/apis/kubedirector.bluedata.io/v1alpha1/zz_generated.openapi.go
```diff
@@ -0,0 +1,227 @@
...
+					"status": {
+						SchemaProps: spec.SchemaProps{
+							Ref: ref("github.com/bluek8s/kubedirector/pkg/apis/kubedirector.bluedata.io/v1alpha1.KubeDirectorClusterStatus"),
+						},
+					},
+				},
+			},
+		},
+		Dependencies: []string{
+			"github.com/bluek8s/kubedirector/pkg/apis/kubedirector.bluedata.io/v1alpha1.KubeDirectorClusterSpec", "github.com/bluek8s/kubedirector/pkg/apis/kubedirector.bluedata.io/v1alpha1.KubeDirectorClusterStatus", "k8s.io/apimachinery/pkg/apis/meta/v1.ObjectMeta"},
```
swamibluedata: Can we use a constant for this. Used at several places.

github.com/bluek8s/kubedirector/pkg/apis/kubedirector.bluedata.io

Makes it easier when moving to kubedirectorplus

pkg/apis/kubedirector.bluedata.io/v1alpha1/zz_generated.openapi.go
```diff
@@ -0,0 +1,227 @@
...
+					"status": {
+						SchemaProps: spec.SchemaProps{
+							Ref: ref("github.com/bluek8s/kubedirector/pkg/apis/kubedirector.bluedata.io/v1alpha1.KubeDirectorConfigStatus"),
+						},
+					},
+				},
+			},
+		},
+		Dependencies: []string{
+			"github.com/bluek8s/kubedirector/pkg/apis/kubedirector.bluedata.io/v1alpha1.KubeDirectorConfigSpec", "github.com/bluek8s/kubedirector/pkg/apis/kubedirector.bluedata.io/v1alpha1.KubeDirectorConfigStatus", "k8s.io/apimachinery/pkg/apis/meta/v1.ObjectMeta"},
```
swamibluedata: similar comment about constant

swamibluedata:
One general comment on the client object. This seems like an ideal candidate for a global client object. 
This way we don't need to be passing down through all the functions. This is how we were using it before

Makefile
don-milos: Ok, I modified the operator.yaml and removed most of the sed stuff, except for the REPLACE_IMAGE line, as you suggested.

Makefile
don-milos: expr manpage:

       Exit status is 0 if EXPRESSION is neither null nor 0, 1 if EXPRESSION is null or 0, 2 if EXPRESSION is syntactically invalid, and 3 if an error occurred.

When this expression went to zero, " retries=`expr $$retries - 1`", the exit status of the expr command was 1 and the make command exited because it thought there was a command failure.

swamibluedata:
One general comment on the client object. This seems like an ideal candidate for a global client object. 
This way we don't need to be passing down through all the functions. This is how we were using it before

cmd/manager/main.go
don-milos: I fixed the Makefile compile target and will migrate our code to the new log implementation.

Makefile
don-milos: There is nothing in our Makefile that invokes build.sh or docker_build.sh so I'm assuming it's the operator-sdk build command. I moved the scripts to build/bin and the build still works.

cmd/manager/main.go
don-milos: Regarding the new log implementation, each Reconcile request creates a new logger object that will tag all log messages with the request's namespace and name:
	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
This eliminates the need for some of the existing shared.logger code that takes the KubeDirectorCluster namespace/name and prepends them to all of the log messages. This also makes the shared/logger code a little more generic because the log message generation it isn't tied to KubeDirectorCluster.
The only downside is that I will need to pass this new logger object around in a majority of all  of the function calls. It would have been nice if I could just replace the KubeDirectorCluster parameter  in the shared/logger APIs with a new logger parameter but  the  KubeDirectorCluster parameter has to be preserved because it is still needed for the event recorder.
We should probably talk offline about possible solutions.

deploy/kubedirector/kubedirector_v1alpha1_kubedirectorcluster_crd.yaml
don-milos: The CR yaml files were created by the operator SDK. We don't really need them so I will delete them.

don-milos: I'm glad you raised the size issue. I had to temporarily comment out the requirement for the size parameter to make progress. I believe the intent is that if (and only if) storage is specified then the size is required. If storage is not specified then the size is not required. I was getting a validation error that size was required even though the KubeDirectorCluster I was trying to create had no storage. I could use some help in debugging this issue.

deploy/kubedirector/kubedirector_v1alpha1_kubedirectorcluster_crd.yaml
don-milos: Those lines were added by the operator-sdk. We curently don't have any KubeDirectorCluster subresources so I will remove them.

deploy/role.yaml
don-milos: The role.yaml file has been deleted.

deploy/service_account.yaml
don-milos: The deploy/service_account.yaml file has been deleted.

version/version.go
don-milos: Done.

tools.go
don-milos: It was created by the operator-sdk. Since it's empty I'll delete it.

don-milos: Actually, I was thinking of a different file. It was created by the operator-sdk. It just imports packages and I don't really know what it's for.

pkg/controller/kubedirectorconfig/doc.go
don-milos: Done.

cmd/manager/main.go
don-milos: Done. Please let me know if the issue that I created is ok.

pkg/controller/kubedirectorcluster/util.go
don-milos: We definitely need to discuss this offline. Once thing I will add though is that we're not actually fetching these objects from Kubernetes all the time. The new framework uses a split client that caches the objects and uses the cache a lot of the time.

pkg/executor/service.go
don-milos: Done.

cmd/manager/main.go
don-milos: Done.

cmd/manager/main.go
don-milos: c, err := controller.New("kubedirectorcluster-controller", mgr, controller.Options{Reconciler: r})
	if err != nil {
		return err
	}

	// Watch for changes to primary resource KubeDirectorCluster
	err = c.Watch(&source.Kind{Type: &v1alpha1.KubeDirectorCluster{}}, &handler.EnqueueRequestForObject{})
	if err != nil {
		return err
	}

When we create the controller and configure the watch we don't specify a namespace so I'm assuming we watch for objects in all namespaces based on their Group/Version/Kind.

cmd/manager/main.go
don-milos: Sorry, I'm not seeing it. What should the error message be?

pkg/apis/kubedirector.bluedata.io/v1alpha1/kubedirectorapp_types.go
don-milos: Done.

cmd/manager/main.go
don-milos: I checked and the Service object is cleaned up by the current Makefile.

pkg/controller/kubedirectorcluster/util.go
don-milos: One more piece of info from the K8s documentation, "A split client reads (Get and List) from the Cache and writes (Create, Update, Delete) to the API server. Reading from the Cache significantly reduces request load on the API server; as long as the Cache is updated by the API server, read operations are eventually consistent."

cmd/manager/main.go
joel-bluedata: My mistake!

pkg/controller/kubedirectorconfig/kubedirectorconfig_controller.go
don-milos: From the delete testing I've done I believe that's the case. Reconcile got called twice on every delete, once where the Get succeeded and then when the Get returned IsNotFound.

build/Dockerfile
don-milos: I ran into issues trying to use the ubi-minimal image. I'm on CentOS and needed certificates to access the image. I had trouble generating the certificates so I fell back to using the alpine image.

Just let me know what image you guys decide on and I'll get it to work.

joel-bluedata: Interesting. I just tried this command manually and it worked for me on my Mac laptop:
```shell
docker pull registry.access.redhat.com/ubi7/ubi-minimal:latest
```

What step of the process was failing for you, and what was the error?

(I mean Alpine is probably fine but I can imagine we might get dinged for not using UBI, so I'm interested in seeing if that's a possibility for us or not.)

joel-bluedata: Sorry, got distracted by a different bug. Anyway it fails for me too on CentOS, but not in any way related to certs (apparently):
```shell
[sources]$ docker pull registry.access.redhat.com/ubi7/ubi-minimal:latest
latest: Pulling from registry.access.redhat.com/ubi7/ubi-minimal
b54017d3c94f: Pulling fs layer
b526fa5b30f2: Pulling fs layer
Pulling repository registry.access.redhat.com/ubi7/ubi-minimal
Error: image ubi7/ubi-minimal:latest not found
```

I did come across this other bug report which does mention certs and seems to conclude "Docker version too old": https://github.com/operator-framework/operator-sdk/issues/1486

FWIW the Docker version in my CentOS VM is super old (1.7.1) while my Mac version is fairly recent (Docker Engine Community 18.09.2). What Docker version are you rolling with?

don-milos: Here's the error I'm seeing:

Sending build context to Docker daemon 320.8 MB
Step 1/6 : FROM registry.access.redhat.com/ubi7/ubi-minimal:latest
Trying to pull repository registry.access.redhat.com/ubi7/ubi-minimal ...
open /etc/docker/certs.d/registry.access.redhat.com/redhat-ca.crt: no such file or directory

deploy/kubedirector/kubedirector_v1alpha1_kubedirectorcluster_crd.yaml
don-milos: After further testing I've found that I had to add these lines back. I was unable to create KubeDirectorCluster's when they were deleted.

pkg/controller/kubedirectorcluster/cluster.go
don-milos: I made the changes you recommended above and tested them. I didn't see any issues so I'm going to check them in.

pkg/controller/kubedirectorcluster/cluster.go
don-milos: Done.

pkg/controller/kubedirectorcluster/cluster.go
don-milos: Done.

pkg/controller/kubedirectorcluster/cluster.go
don-milos: Done.

pkg/controller/kubedirectorcluster/kubedirectorcluster_controller.go
don-milos: I decided to move the  invocation of removeClusterAppReference to handleFinalizers() to, as you said, keep all of the cleanup activities bundled together. I think that's cleaner and easier to understand.

I don't think we have to move the call to  ensureClusterAppReference after the early exit triggered by handleFinalizers. Currently we'll  call ensureClusterAppReference at the beginning of syncCluster, then we'll call removeClusterAppReference from handleFinalizers and then exit early from syncCluster. Is there an issue with that sequence?

don-milos:
> One more thing. In the old framework we could specify the resyncPeriod for watching a particular CR, so that we could control how frequently our handler might get triggered. Is there an equivalent in the new framework?

Good catch. That was on my todo list but I didn't get to it. I've checked in the changes but I have a couple of questions.

I currently have the polling cycle set to 30 secs for both the Config and Cluster CRs.

1. Is that polling cycle OK for the cluster CRs?

2. Can we not poll at all for the config CRs? There's no status to check and we can just reconcile when a change is posted.

Makefile
don-milos: These changes for relaxed security are only to allow a debugger to attach to kubedirector and I've successfully done that so I don't think this change is necessary. At least not at the moment.

Makefile
don-milos: I used NodePort to make debugging convenient. This is the port used to connect to the debugger process within the container. In the debugger UI you need to specify an IP address and a port to connect to. If we used the NodePort we would have to use the container IP address and that's always changing. By using the HostPort we can always use localhost as the IP address, assuming the UI is running on the same node as the container.

deploy/operator.yaml
don-milos: The file already looks like that so I'm not sure what you want me to change?

pkg/controller/kubedirectorcluster/kubedirectorcluster_controller.go
don-milos: I just read my previous reply and I feel I need to expand it a little.

I don't think we have to move the call to ensureClusterAppReference after the early exit triggered by handleFinalizers. Currently we'll call ensureClusterAppReference at the beginning of syncCluster that potentially adds an app reference for the KD cluster, then we'll call removeClusterAppReference from handleFinalizers that removes the KD Cluster's apps reference, and then we exit early from syncCluster. Bottom line is that the reference is always removed when the KD cluster is deleted which is what we want.

Is there an issue with that sequence?

deploy/operator.yaml
joel-bluedata: I think Swami made this comment before your most recent changes.

pkg/executor/cluster.go
don-milos: Until a stabile version of controller-runtime supports patch I've implemented the follow the following from Joel's recommendations:
-	Start with the given input object as the object-to-modify.
-	Apply the given modify function to the object-to-modify.
-	Try to PUT that object.
-	If that succeeds, we can return.
-	If it failed for reasons other than a resourceVersion conflict just return an error.
-	If there was a resourceVersion conflict, do a GET and modify a more recent version of the object.
-	If that fails return the error, return nil on success.
I decided to not add retry logic for the updates since no retry logic existed previously. It can easily be added later if deemed necessary. Retry logic already existed for cluster UpdateStatus and is still in place.

pkg/executor/cluster.go
don-milos: See reply to lengthy PATCH comment above.

pkg/executor/cluster.go
don-milos: See reply to lengthy PATCH comment above.

Makefile
don-milos: This reply is regarding Joel's comment about polling intervals above. I tried a quoted reply but it's at the bottom of this thread so this is a duplicate.

Good catch. That was on my todo list but I didn't get to it. I've checked in the changes but I have a couple of questions.

I currently have the polling cycle set to 30 secs for both the Config and Cluster CRs.

1. Is that polling cycle OK for the cluster CRs?

2. Can we not poll at all for the config CRs? There's no status to check and we can just reconcile when a change is posted.

pkg/apis/kubedirector.bluedata.io/v1alpha1/zz_generated.openapi.go
joel-bluedata: Since this is generated code I think this particular spot is out of our hands. I'll talk to you more about this offline.

cmd/manager/main.go
don-milos: I discussed the log issue with Joel yesterday offline and we decided to go ahead with the logger change. The logger setup at the beginning of each reconciliation request will be used consistently throughout the code.

build/Dockerfile
joel-bluedata: Yeah that seems to be the "docker version too old" thing mentioned above. What Docker version do you have?

don-milos: docker-client-1.13.1-96.gitb2f74b2.el7.centos.x86_64
docker-common-1.13.1-96.gitb2f74b2.el7.centos.x86_64
docker-1.13.1-96.gitb2f74b2.el7.centos.x86_64

cmd/manager/main.go
joel-bluedata: Hmm. Do you know how? There's no explicit delete for that service in the Makefile I think. Is it created with an owner ref?

cmd/manager/main.go
joel-bluedata: Hmm. Do you know what the namespace given to the manager is used for then?

Anyway this will be easy enough to test/verify.

pkg/controller/kubedirectorcluster/kubedirectorcluster_controller.go
joel-bluedata: Seems fine yeah. Trying to remember if there was something else that bothered me about that beyond the potential redundancy... I guess not?

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -91,60 +80,60 @@ func syncCluster(
...
 				time.Sleep(wait)
 			}
 		}
 	}()
 
-	// Ignore stale poll-driven handler for a resource we have since
-	// updated. Also for a new CR just update the status state/gen.
-	shouldProcessCR := handleStatusGen(cr, handler)
-
 	// Regardless of whether the status gen is as expected, make sure the CR
```
joel-bluedata: Should remove the "regardless" clause here.

tools.go
joel-bluedata: I believe this is only used for stuff that is built using "go modules". Cf. https://github.com/golang/go/issues/25922

We haven't switched to using modules yet, but Tom K. did take a whack at allowing us to maintain the necessary metadata (go.mod/sum) in the original KubeDirector code. So perhaps we should leave this file here as well. I do wonder why the SDK framework generated it... maybe it was left over from a previous try at making the operator skeleton using modules rather than dep?

pkg/executor/cluster.go
joel-bluedata: Previously we didn't need retry logic in our code because we were not vulnerable to resourceVersion conflicts.

But it may be the case, at least for some of these PATCHes, that the right thing to do is fail and bail out of the reconcile handler (or at least that portion of the handler, if subsequent operations don't depend on the PATCH being done) and then let the next handler invocation try again. Basically move the retry logic out of our code and into the SDK framework. The advantage of that approach would be that we don't need to have some cap on number of retries and logic about what to do when that cap is hit.

We'll need to review the current PATCH invocations to see what will happen if we immediately bail out on error.

Makefile
joel-bluedata: 1. Yes that's what we had previously.

2. I think that's what the previous resyncPeriod of 0 meant for the config CRs.

build/Dockerfile
joel-bluedata: OK so FWIW, here's how I got that to work.

Those certificates are supposed to live in a certain folder. That folder is supplied by the python-rhsm-certificates package. However on my test host (EC2 VM), it's also supplied by the currently installed subscription-manager-rhsm-certificates package, which just puts an empty directory there. That package is marked as obsoleting python-rhsm-certificates.

So I tried this:
```shell
rpm -e --nodeps subscription-manager-rhsm-certificates
yum --setopt=obsoletes=0 install python-rhsm-certificates
```
Then I could do the docker pull.

(I had to use --nodeps so that dependent packages like Docker wouldn't get yanked out.)

This is not a super great solution but maybe it helps point in the right direction? And if your current CentOS environment does NOT have the subscription-manager-rhsm-certificates package installed, then this may be a super easy fix and all you need to do is that yum install command above.

deploy/kubedirector/rbac-default.yaml
```diff
@@ -39,6 +39,7 @@ rules:
   - apps
   resources:
   - deployments
+  - replicasets
```
joel-bluedata: Was this something that came from the generated SDK framework somehow, or did you find that you needed to add this?

pkg/controller/kubedirectorcluster/util.go
don-milos: I moved the global data out of the KubeDirectorCluster controller to shared module so that it's cached and accessible from all the modules like it was before.

pkg/controller/kubedirectorcluster/kubedirectorcluster_controller.go
don-milos: Removed the hack by moving he global data out of the KubeDirectorCluster controller to shared module.

pkg/shared/globalconfig.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+var (
+	globalConfig     *kdv1.KubeDirectorConfig
+	globalConfigLock sync.RWMutex
+)
+
+// getNativeSystemdSupport extracts the flag definition from the
+// globalConfig CR data if present, otherwise returns false
+func GetNativeSystemdSupport() bool {
```
joel-bluedata: Seems like these read-accessor functions should also lock the globalConfigLock. I don't think that pointer-variable update is guaranteed to be atomic in the Go spec.

joel-bluedata:
BTW I notice that the Travis CI build is failing with this error message:
```
build github.com/bluek8s/kubedirector/cmd/manager: cannot load github.com/Azure/go-autorest/autorest: ambiguous import: found github.com/Azure/go-autorest/autorest in multiple modules:
	github.com/Azure/go-autorest v11.7.0+incompatible (/home/travis/gopath/pkg/mod/github.com/!azure/go-autorest@v11.7.0+incompatible/autorest)
	github.com/Azure/go-autorest/autorest v0.5.0 (/home/travis/gopath/pkg/mod/github.com/!azure/go-autorest/autorest@v0.5.0)
Makefile:266: recipe for target 'compile' failed
```

Kinda weird. Does "make compile" work for you locally?

cmd/manager/main.go
joel-bluedata: I talked with Swami a bit more about this and took another look at the code...

Previously we got this GroupVersionKind from the "kd" object which was a Deployment. The "Kind" therefore would have been "Deployment". So I think it should be "Deployment" here, and not "KubeDirectorCluster".

I'm a little surprised that "KubeDirectorCluster" is working here. This is used to create an owner ref that makes various infrastructure objects (Secret named kubedirector-validator-secret, Service named kubedirector-validator, and Webhook named kubedirector-webhook) will get auto-deleted if the kubedirector Deployment is deleted. Might be interesting to double-check whether that is currently happening.

build/Dockerfile
don-milos: Your workaround worked for me. I tested with ubi-minimal as the base image and it seems to work ok.

pkg/shared/globalconfig.go
```diff
@@ -0,0 +1,72 @@
...
+	globalConfigLock.Lock()
```
joel-bluedata: Can we use RLock instead of Lock for read-only stuff?

joel-bluedata:
Just FYI my plan is to still land this on a branch this evening and move any unresolved issues to tracked Issues on GitHub. Any remaining work on this branch today should focus on low-hanging fruit.

build/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM registry.access.redhat.com/ubi7/ubi-minimal:latest
+ENV OPERATOR=/usr/local/bin/kubedirector
+
+# Allow delve to run on Alpine based containers.
+# RUN apk add --no-cache libc6-compat
```
joel-bluedata: Probably should remove these comments if they really are Alpine-specific.

cmd/manager/main.go
don-milos: I'm still investigating this one. I tried hardocoding the values used by the original KubeDirector code (i.e. Group:   "apps", Version: "v1", Kind: "Deployment") and the webhooks didn't start properly.  I also mentioning in an email exchange that we had that “kd.GroupVersionKind()” is returning empty GVK strings.

pkg/controller/kubedirectorcluster/cluster.go
don-milos: -       // Regardless of whether the status gen is as expected, make sure the CR
-       // finalizers are as we want them. We use a finalizer to prevent races
-       // between polled CR updates and CR deletion.
+       // We use a finalizer to ensure that only KubeDirector updates status

don-milos: Not sure what happened there. The new comment reads, "We use a finalizer to ensure that only KubeDirector updates status"

joel-bluedata:
List of things that will be converted to GitHub issues tonight (unless there's some in-flight change or response that lands soon addressing it):
* What deletes the metrics service object on cleanup?
* Why is the storage.size KDC property always required in Don's setup?
* Why is a "status" subresource required to be declared in Don's setup?
* Does KD correctly watch all namespaces for KDCluster? What is the namespace passed to the manager used for?
* Does KD correctly only watch its OWN namespace for KDConfig?
* Fixes for object-update behavior (e.g. handle necessary retries across multiple handler invocations, instead of hardcoded single-retry within handler).
* Set polling interval to 30 seconds for KDCluster. Don't poll for KDConfig.
* Set correct owner ref for created secret/service/webhook and verify they go away when KD deployment is deleted.
* Management of the K8s client (shared global vs. passed around).
* Remove replicasets from the RBAC, or determine why that is not possible.
* Fix Travis CI build.

don-milos:
> Set polling interval to 30 seconds for KDCluster. Don't poll for KDConfig.
I checked in a fix for this issue earlier this week.

joel-bluedata:
Issues filed: https://github.com/bluek8s/kubedirector/labels/Project%3A%20Operator%20SDK

Issue #181, opened 2019-07-27
remove replicasets from the RBAC


joel-bluedata:
We really shouldn't need to have that in there.

If removing it causes an error, we need to track down what operation is claiming to need that privilege.

don-milos:
I removed replicasets from deploy/kubedirector/rbac-default.yaml and was unable to reproduce the error that I got before. It's safe to remove it.

don-milos:
My bad. After further testing I discovered there is an error when I delete replicasets from kubedirector's RBAC privileges:

{"level":"info","ts":1564415451.8044612,"logger":"kubedirector","msg":"failed to initialize service object for metrics: replicasets.apps \"kubedirector-7f97d5959b\" is forbidden: User \"system:serviceaccount:default:kubedirector\" cannot get resource \"replicasets\" in API group \"apps\" in the namespace \"default\""}

don-milos:
# kubectl describe deployment kubedirector
Name:                   kubedirector
Namespace:              default
CreationTimestamp:      Mon, 29 Jul 2019 11:41:28 -0400
Labels:                 <none>
Annotations:            deployment.kubernetes.io/revision: 1
Selector:               name=kubedirector
Replicas:               1 desired | 1 updated | 1 total | 1 available | 0 unavailable
StrategyType:           RollingUpdate
MinReadySeconds:        0
RollingUpdateStrategy:  25% max unavailable, 25% max surge
Pod Template:
  Labels:           name=kubedirector
  Service Account:  kubedirector
  Containers:
   kubedirector:
    Image:      donmilos1/kubedirector:unstable
    Port:       60000/TCP
    Host Port:  0/TCP
    Command:
      /bin/sh
    Args:
      -c
      mkfifo /tmp/fifo; (/usr/local/bin/kubedirector &> /tmp/fifo) & while true; do cat /tmp/fifo; done
    Environment:
      MY_NAMESPACE:      (v1:metadata.namespace)
      WATCH_NAMESPACE:   (v1:metadata.namespace)
      POD_NAME:          (v1:metadata.name)
      OPERATOR_NAME:    kubedirector
    Mounts:             <none>
  Volumes:              <none>
Conditions:
  Type           Status  Reason
  ----           ------  ------
  Available      True    MinimumReplicasAvailable
  Progressing    True    NewReplicaSetAvailable
OldReplicaSets:  <none>
NewReplicaSet:   kubedirector-7f97d5959b (1/1 replicas created)
Events:
[...]

PR #185, opened 2019-07-29
Remove replicasets from the RBAC, issue #181

don-milos:
I removed replicasets from deploy/kubedirector/rbac-default.yaml and was unable to reproduce the error that I got before. It's safe to remove it.

### DIFF
--- a/deploy/kubedirector/rbac-default.yaml
+++ b/deploy/kubedirector/rbac-default.yaml
@@ -39,7 +39,6 @@ rules:
   - apps
   resources:
   - deployments
-  - replicasets
   verbs:
   - "get"
 - apiGroups:


don-milos:
My bad. After further testing I discovered there is an error when I delete replicasets from kubedirector's RBAC privileges:

{"level":"info","ts":1564415451.8044612,"logger":"kubedirector","msg":"failed to initialize service object for metrics: replicasets.apps \"kubedirector-7f97d5959b\" is forbidden: User \"system:serviceaccount:default:kubedirector\" cannot get resource \"replicasets\" in API group \"apps\" in the namespace \"default\""}

Closing this pull request.

PR #186, opened 2019-07-30
Fix GVK supplied to create owner reference for the secret/service/web…

don-milos:
This issue explains why the GVK is empty in the Deployment object that we get back from client-go: kubernetes/client-go#541 .

I will modify the code to use the correct GVK.


joel-bluedata:
Good find with that issue!

Issue #180, opened 2019-07-27
client access via global handle rather than arg passed everywhere


joel-bluedata:
On new operator SDK branch.

don-milos:
https://github.com/bluek8s/kubedirector/pull/191

PR #191, opened 2019-08-06
Access client via global handle rather than arg passed everywhere.

don-milos:
Address issue #180

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -16,9 +16,11 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"github.com/bluek8s/kubedirector/pkg/observer"
+	"github.com/bluek8s/kubedirector/pkg/shared"
 	"github.com/bluek8s/kubedirector/pkg/validator"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"os"
@@ -122,7 +124,7 @@ func main() {
 
 	// Setup Scheme for all resources
 	if err := apis.AddToScheme(mgr.GetScheme()); err != nil {
-		log.Error(err, "")
+		log.Error(err, "failed to add KubeDirector CRs to scheme")
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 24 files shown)


don-milos:
The Travis build is failing because GO111MODULE="on" is set in .travis.yaml. That was deleted in a prior checkin.

cmd/manager/main.go
```diff
@@ -138,22 +140,43 @@ func main() {
...
 
+	// Initialize the shared K8s Client that is used for all K8s CRUD
+	// See https://github.com/bluek8s/kubedirector/issues/173
+	// Since we are not using the manager's webhook framework and are setting
+	// up our own validation server we need to use a temporary client to do that
+	// because the manager's client won't work before mgr.Start() is called and
+	// the the manager's cache is initialized. Once the cache is initialized we
+	// can start using the manager's split client.
+	shared.Client, err = client.New(
+		mgr.GetConfig(),
```
joel-bluedata: Couple of things bothering me here. Obviously this is working currently, but as a matter of (I guess) form:
* Since we're initializing a var in the shared package, we should probably use a function in that package to do it.
* We have a couple of different "configs" flying around the codebase now.

W.r.t. that last point... there's a shared.ClientConfig that is initialized from `rest.InClusterConfig()`, which seems to only be used by shared.execCommand. There's also this config used here by the manager and by shared.Client which is initialized from `config.GetConfig()`.

`config.GetConfig()` does fall back to `rest.InClusterConfig()` internally if there's no command-line argument or environment variable telling it to do otherwise, so we're probably ending up with the same config in both cases here. But it seems like we should unify this.

My initial reflex is that the code initializing shared.ClientConfig should use `config.GetConfig()`, and then everyone (including the manager and shared.Client) should use shared.ClientConfig.

Then coming back to my first quibble... it would almost be the case that shared.Client could be properly initialized inside the init function of client.go, and we wouldn't have to do anything explicitly from this code. However there's the thing about needing `mgr.getScheme()` for the client options. Hmm

It's not the end of the world if we need to call some shared-package function from here and pass in the scheme as an argument, to initialize shared.Client. Although this may be related to your previous questions about building an appropriate scheme? Maybe the shared-package init should also be where a scheme is built that is then used both by the manager as well as by this client.

don-milos: - Since we're initializing a var in the shared package, we should probably use a function in that package to do it.

I agree and will make that change.

Regarding your other points, I agree we shouldn't have multiple "configs" floating around but I would prefer to use the Config and Scheme provided by the manager. How about I add a function call in main() with the manager as an argument and the shared package uses manager to initialize the shared Config and Client? For example, shared.Init(mgr)? That will remove all the shared package business logic from main.

joel-bluedata: The Config provided by the manager is presumably the same one we passed into it, which is from config.GetConfig() right? I think we can do config.GetConfig() once somewhere and then give that to whoever wants it. I would kiiind of like to keep manager-awareness out of the shared-package code although I don't feel super strongly about that.

don-milos:
I think I read too much into your previous comments. I agree with everything you said and pushed some changes to address them. Thanks!

Issue #194, opened 2019-08-20
Migrate Client().Update() calls back to Patch() calls


don-milos:
When KubeDirector was migrated from operator-sdk version v0.0.6 to v0.8.1 the calls to sdk.Patch() were replaced with calls to Client().Update() because the Patch() API was removed. The Update() call replaces the entire resource object instead of a few selected fields in the object. Also, the objects now contain a ResourceVersion and if the update object is not at the latest ResourceVersion then the update fails with an IsConflict error with a message to try again with an up to date object. To recover from this error additional code was added to fetch the most recent version of the object, patch it, and then try the update again with the patched object.

The Patch() API is supposed to reintroduced in the next release of the K8s controller-runtime library. When a stable release with the Patch() API is available we need to port the KubeDirector back to using patch instead of update and remove all the retry logic from the code.

joel-bluedata:
We can keep an eye on this file to see what version of controller-runtime is currently being used by the SDK: https://github.com/operator-framework/operator-sdk/blob/master/internal/pkg/scaffold/gopkgtoml.go

joel-bluedata:
Next operator SDK release after v0.10.0 should using controller runtime 0.2.0.

joel-bluedata:
Yep, releases v0.11.0 and later of the operator SDK use the controller-runtime that has PATCH support.

joel-bluedata:
However when we update past v0.8.2, we'll have to finally transition from using dep to using Go modules instead, which, sigh. There's always something!

Created issue #222 for that.

joel-bluedata:
Starting work on porting to new operator SDK in https://github.com/joel-bluedata/kubedirector/tree/sdk

joel-bluedata:
Using new SDK now.

joel-bluedata:
Not a blocker for 0.5.0, but it would be nice to tackle at least some of these so I will leave this on the 0.5.0 milestone for now.

Issue #178, opened 2019-07-27
harden object-update behavior


joel-bluedata:
On the new operator SDK branch, we can't do a PATCH through the client because there's not a stable version of controller-runtime that supports PATCH yet. So we have to do read-modify write in cases where we were previously doing a PATCH. But that is racy and could fail with a resourceVersion conflict if some other client modifies the object around the same time.

We need to gracefully handle failure and retry. The most custom-controller-ish way to handle this I think is to bail out of the reconciliation handler so that it can try again at the next polling interval. We would need to take a close survey of the code locations that do this kind of update to see how to best fail our way out of the handler from that point, and whether that naturally leaves us in a consistent state that can be picked up again by the next reconciliation pass.

The main exception to this would be the update of the KDC status, which is required to succeed before exiting the handler (because that's where we keep important persistent state that drives reconciliation). In that case if the update fails we have to keep retrying, with no exit conditions other than success or object-not-found.

don-milos:
The code in controller/kubedirector_cluster.go/Reconcile() does just what you describe. On error the reconcile.result contains the error so the request is requeued to be tried again.

Also, I believe the controller/kubedirectorcluster_controller/cluster.go/syncCluster() function already retries (forever) the UpdateStatus request until success or object-not-found before it will exit.

joel-bluedata:
Yeah I think the status update func in syncCluster is OK, I just wanted to make a note that we should not break its retry-forever behavior as we mess with PATCH implementations.

As for the other spots, I'm talking about the lower-level code that does the object update such as in executor.UpdateStatus or RemoveFinalizer etc. (basically spots where errors.IsConflict is currently being checked). We have code that retries once and then returns an error if that single retry fails.

My original proposal was not for a single retry, but rather to retry until success. My revised proposal however was that we should have the courage of our convictions (about our operator/reconciliation model) and not even retry once. If there's an error, we should exit the handler cleanly and rely on the next handler go-around to try again. Need to examine these spots and make sure that is indeed safe to do.

don-milos:
There is a pre-existing comment at lines 52-53 in syncCluster() to note that we should try forever:

        // Write back the status. Don't exit this reconciler until we
        // succeed (will block other reconcilers for this resource).

Even though executor/cluster.go/UpdateStatus() only tries twice to update the status we've already verified that the loop in syncCluster() will try to update the status forever, by calling updateStatus(), until it succeeds or the resource is deleted.

[...]

PR #195, opened 2019-08-20
GitHub issue #178: harden object-update behavior

don-milos:
Simplified the update retry code and added comments to reference the newly created GitHub issue on this topic.

https://github.com/bluek8s/kubedirector/issues/194

### DIFF
--- a/pkg/controller/kubedirectorcluster/cluster.go
+++ b/pkg/controller/kubedirectorcluster/cluster.go
@@ -15,17 +15,16 @@
 package kubedirectorcluster
 
 import (
-	"github.com/go-logr/logr"
-	"reflect"
-	"time"
-
 	kdv1 "github.com/bluek8s/kubedirector/pkg/apis/kubedirector.bluedata.io/v1alpha1"
 	"github.com/bluek8s/kubedirector/pkg/catalog"
 	"github.com/bluek8s/kubedirector/pkg/executor"
 	"github.com/bluek8s/kubedirector/pkg/observer"
 	"github.com/bluek8s/kubedirector/pkg/shared"
+	"github.com/go-logr/logr"
 	"github.com/google/uuid"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"reflect"
+	"time"
 )
 
 // syncCluster runs the reconciliation logic. It is invoked because of a
@@ -35,9 +34,6 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 	reqLogger logr.Logger,
 	cr *kdv1.KubeDirectorCluster,
 ) error {
-
-	// Make sure this cluster marks a reference to its app.
-	shared.EnsureClusterAppReference(cr.Namespace, cr.Name, cr.Spec.AppID)
 
 	// Make sure we have a Status object to work with.
 	if cr.Status == nil {
@@ -74,6 +70,20 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 					if currentCluster.DeletionTimestamp != nil {
 						return
 					}
+
+					// See GitHub issue #194: Migrate Client().Update() calls
+					// back to Patch() calls.
+					// https://github.com/bluek8s/kubedirector/issues/194
+					if errors.IsConflict(updateErr) {
+						// If the update failed with a ResourceVersion
+						// conflict then we need to use the current
+						// version of the cluster. Otherwise, the status
+						// update will never succeed and this loop will
+						// never terminate.
+						currentCluster.Status = cr.Status
+						*cr = *currentCluster
+						continue
+					}
 				}
 				if wait < maxWait {
 					wait = wait * 2
@@ -91,18 +101,22 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 		}
 	}()
 
-	// We use a finalizer to ensure that only KubeDirector updates status
+	// For a new CR just update the status state/gen.
+	shouldProcessCR := r.handleNewCluster(reqLogger, cr)
+	if !shouldProcessCR {
+		return nil
+	}
+
+	// Make sure this cluster marks a reference to its app.
+	shared.EnsureClusterAppReference(cr.Namespace, cr.Name, cr.Spec.AppID)
+
+	// We use a finalizer to maintain KubeDirector state consistency;
+	// e.g. ClusterAppReference's and StatusGen's.
 	doExit, finalizerErr := r.handleFinalizers(reqLogger, cr)
 	if finalizerErr != nil {
 		return finalizerErr
 	}
 	if doExit {
-		return nil
-	}
-
-	// For a new CR just update the status state/gen.
-	shouldProcessCR := r.handleNewCluster(reqLogger, cr)
-	if !shouldProcessCR {
 		return nil
 	}
 
--- a/pkg/controller/kubedirectorcluster/cluster.go
+++ b/pkg/controller/kubedirectorcluster/cluster.go
@@ -83,9 +83,6 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 					if currentCluster.DeletionTimestamp != nil {
 						return
 					}
-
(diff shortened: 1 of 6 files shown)


pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -74,6 +70,20 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 					if currentCluster.DeletionTimestamp != nil {
 						return
 					}
+
+					// See GitHub issue #194: Migrate Client().Update() calls
+					// back to Patch() calls.
```
joel-bluedata: Since we're now using a subresource for Status, I don't think the patch issue is relevant here.

Or that a resourceVersion conflict is possible (since no one other than us can update the status)? Although I guess I'm just assuming that the main-resource resourceVersion is irrelevant for the subresource. Have you observed a resourceVersion conflict here or is this block just being cautious?

pkg/executor/statefulset.go
```diff
@@ -65,6 +65,11 @@ func UpdateStatefulSetReplicas(
 	if err == nil {
 		return nil
 	}
+
+	// See GitHub issue #194: Migrate Client().Update() calls
+	// back to Patch() calls.
+	// https://github.com/bluek8s/kubedirector/issues/194
```
joel-bluedata: We also need a comment like this in executor.UpdatePodService.

pkg/executor/cluster.go
```diff
@@ -94,59 +42,18 @@ func RemoveFinalizer(
...
-				reqLogger,
-				err,
-				cr,
-				shared.EventReasonCluster,
-				"failed to remove finalizer",
-			)
-		}
-	}
-	return err
+	return shared.Client().Update(context.TODO(), cr)
```
joel-bluedata: Should have the comment here about moving to PATCH when available.

pkg/executor/cluster.go
```diff
@@ -212,22 +70,10 @@ func addFinalizer(cr *kdv1.KubeDirectorCluster) bool {
...
-	found := false
-	for i, f := range cr.Finalizers {
-		if f == finalizerID {
-			cr.Finalizers = append(cr.Finalizers[:i], cr.Finalizers[i+1:]...)
-			found = true
-			break
-		}
-	}
-	return found
+	return shared.Client().Update(context.TODO(), cr)
```
joel-bluedata: Should have the comment here about moving to PATCH when available.

pkg/controller/kubedirectorcluster/cluster.go
don-milos: The ResourceVersion field is shared between the Spec or Status and is updated whenever either is modified. I have observed the for loop failing forever so checking for the conflict error here is necessary.

pkg/executor/statefulset.go
don-milos: Done.

pkg/executor/cluster.go
don-milos: Done.

pkg/executor/cluster.go
don-milos: Done.

pkg/controller/kubedirectorcluster/cluster.go
joel-bluedata: Got it. OK, the conflict check is necessary then, but I think the comment about using PATCH is not necessary here.

don-milos: Even though we're not calling Update() here I think the comment is still necessary because when we go back to using Patch() I don't think we'll get the conflict error so the conflict check can be removed. Right?

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -91,7 +101,17 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 		}
 	}()
 
-	// We use a finalizer to ensure that only KubeDirector updates status
+	// For a new CR just update the status state/gen.
+	shouldProcessCR := r.handleNewCluster(reqLogger, cr)
+	if !shouldProcessCR {
+		return nil
+	}
```
joel-bluedata: So, hmm. Why this new ordering? Let's chew on this.

One thing that seems like a problem here is that we now *won't* call EnsureClusterAppReference when the cluster is first seen (because shouldProcessCR will be false). 

Also: previously we had intentionally moved handleFinalizers up *before* handleNewCluster, because we needed to move deleteStatusGen into handleFinalizers. Here was the reasoning for that:

-----
...we now have the situation where we could get a poll-triggered handler for a KDC CR that has a deletion timestamp and no finalizer, which has no entry in our status gen cache. We need to make sure that is handled correctly, i.e. ignored

So in the KDC reconciler, we want to make sure we don't do any processing for a CR that has a deletion timestamp, beyond (if necessary) removing its finalizer and removing its status gen from the cache. It looks like handleFinalizers already takes care of that. However, the current ordering of the calls to handleNewCluster (used to be handleStatusGen) and handleFinalizers is unfortunate in that handleNewCluster would re-add the status gen back into the cache. I *believe* the simplest answer to that problem is to switch the order that those functions are called -- always do handleFinalizers first.
-----

Our churn in how we are invoking these functions may be a sign to revisit them a bit.

This are of code is also related to some changes that Swami made in PR #168. In Swami's PR, handleStatusGen on a new cluster needs to be done *before* we try to add a finalizer (for various reasons). So I was thinking that, when I merge in this operator-sdk branch, I would need to split up the handleFinalizers code into two pieces and do things in this order:
* first take care of the `cr.DeletionTimestamp != nil` part of handleFinalizers
* next, bail out of reconcile if `cr.DeletionTimestamp != nil` (is deleted cluster)
* now take care of handleNewCluster (status gen stuff)
* bail out of reconcile if handleNewCluster asks us to  (is new cluster)
* finally take care of the EnsureFinalizer part of handleFinalizers
* bail out of reconcile if EnsureFinalizer failed

Since those are my plans, I'm curious about why you saw a need to reorder handleFinalizers/handleNewCluster in this PR and whether that issue would affect those plans.

Finally, w.r.t. the cluster app reference stuff...

The app reference is mostly about making sure that the app can't go away after we've started working on the cluster. If it goes away in a little race window after admitting the cluster and before we start working on the cluster, fundamentally we only need to make sure that we fail gracefully. And I believe we would.

Going back to Swami's PR here though: with his change, the validator specifically records the namespace of the app CR in the status of the cluster CR. It would probably make sense to do EnsureClusterAppReference *there*, in the validator, when admitting the cluster CR. The only other place we need to do EnsureClusterAppReference would be in handleNewCluster when uptaking an existing cluster after KD restart.

I'm not 100% sure what this means for your PR vs. what I just need to do in the merge.

I can do a Skype call to chat further. On a customer call for a bit longer here, then free after.

don-milos: I took the following comment as the intended behavior, "For a new CR just update the status state/gen." I took that to mean the statusGen would be assigned and updated in the resource before anything else happens. The code was allowing the finalizer update before because of a side effect/bug with the new update code. When I fixed that issue the validator was rejecting the finalizer update. L I can play with order again to see if I can get the finalizers updated first. et's talk when you get a chance.

joel-bluedata: OK. Gonna lunch real quick first.

don-milos: I've been playing with the order and I think I have it the way that you want it. One issue I ran into was that the default Status object was being assigned too early. Since the Status in the cr object was changing the Client().Update() call to add the finalizer was failing with the error that Update() can't update the Status. Please have a look at the new committed code and let me know if it's ok. Thanks ...

pkg/controller/kubedirectorcluster/cluster.go
joel-bluedata: I would think we would keep using Status().Update() here because we'll still be updating the entire status subresource. (We can't patch the status clause of the main resource because that will be ignored, given the existence of the subresource.)

PR #202, opened 2019-09-17
run as non-root user

joel-bluedata:
The ability to run as root is not going to be available to us in some cases, so let's not rely on that.

The changes to pkg/validator/* are for moving our internal webhook port to a non-restricted port (8443). So we don't need to be root to listen there. The external port for the service will still be 443 but that's fine.

The rest of the changes are for actually running as non-root. They broadly involve these things:

* Remove the old build.sh and docker_build.sh scripts since they aren't used.

* Remove the debug target in the Makefile and the (commented-out) debug stuff in the Dockerfile. Those can probably be re-instituted in some form if we want, but Don will probably have to help figure that out and test it.

* In the Dockerfile, set up to run as a non-root user, and store our files in that user's home directory. There's a twist here: on OpenShift we won't know a priori what user ID will be used. I've made the changes in Dockerfile and in the associated entrypoint script to try to account for that in the way recommended by operator-sdk, tweaked to support our "redeploy" shenanigans, but I haven't tested this on OpenShift yet. But it at least works on non-OpenShift and may leave less work for us to do if/when we tackle OpenShift support for real. It's also generally nice to be more congruent with what the operator-sdk does by default.

* Because the (potentially necessary) entrypoint actions are now fairly complicated, it seems like the best approach is to let the Dockerfile's declared entrypoint command handle things (by running the necessary script) and never override it in the deployment YAML. So I removed the command+args from operator.yaml and deployment-prebuilt.yaml.

* Change the definition of configcliSrcFile in kubedirectorcluster/types.go to account for configcli.tgz moving to the kubedirector user's home directory.

* Change the "redeploy" target of the Makefile to account for the new locations and permissions of files.


build/Dockerfile
```diff
@@ -1,22 +1,37 @@
...
+# OpenShift support, make sure they are usable by any user in the root group.
+RUN mkdir -p ${USER_HOME} && \
+    chown ${USER_UID}:0 ${USER_HOME} && \
+    chmod ug=rwx ${USER_HOME}
+COPY build/_output/bin/kubedirector ${USER_HOME}/kubedirector
+COPY build/_output/configcli.tgz ${USER_HOME}/configcli.tgz
+RUN chown ${USER_UID}:0 ${USER_HOME}/kubedirector && \
+    chmod ug=rwx ${USER_HOME}/kubedirector && \
+    chown ${USER_UID}:0 ${USER_HOME}/configcli.tgz && \
+    chmod ug=rw ${USER_HOME}/configcli.tgz
```
swamibluedata: so the dynamic user will always be part of root group? Is that the reason this will work?

build/bin/entrypoint
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh -e
+
+# This is documented here:
+# https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines
+
+if ! whoami &>/dev/null; then
+  if [ -w /etc/passwd ]; then
+    echo "kubedirector:x:$(id -u):$(id -g):kubedirector user:/home/kubedirector:/sbin/nologin" >> /etc/passwd
```
swamibluedata: You can also use $USER_HOME. I see that is being set in Dockerfile
Similarly for $USER_NAME also

build/Dockerfile
joel-bluedata: Yep.

build/bin/entrypoint
joel-bluedata: Ah true. I think that's what the operator-sdk generated examples do. I forgot that those ENV values would be available here too.

build/Dockerfile
```diff
@@ -1,22 +1,37 @@
...
+# we'll get a random user ID at runtime (rather than USER_UID).
+RUN chmod g=u /etc/passwd
 
-# Debugger
-# COPY build/bin/dlv ${DEBUGGER}
-# RUN chmod +x ${DEBUGGER}
+# Install entrypoint script. This will create the user passwd entry and
+# launch the operator.
+COPY build/bin/entrypoint /usr/local/bin/entrypoint
+RUN chown 0:0 /usr/local/bin/entrypoint && \
```
don-milos: Why is the file owned by root instead of ${USER_UID} ?

Also, I know Swami already asked about this but can you elaborate why the group needs to be root?

don-milos:
Sometimes it's necessary to run privileged commands, e.g. the mount command for FsMounts. How are you going to accommodate that? Should this PR include changes to give the kubedirector user some kind of sudo access?

joel-bluedata:
Any privileged commands like that should happen in some other utility pod (e.g. a daemonset) rather than inside the kubedirector pod. If a particular K8s deployment is just dead set against allowing any root-user pods then I think it's a somewhat better story to say they don't get particular feature X rather than that they can't use kubedirector at all. We can talk separately in email or whatever though if you have further concerns along those lines.

build/Dockerfile
joel-bluedata: Just owned by root as a general pattern for the executables that live in /usr/local/bin. It's still broadly executable.

As for the root-group thing...

If the platform deploying this container honors the image's specified user & group, then it will be run with user ID 1001 and group ID 0. On OpenShift however (and perhaps other platforms with similar philosophy), the runner will have some other arbitrary/random user ID but still have group ID 0. Therefore we have to treat group ID as the important thing for accessing our operator-specific files. A little more about that here: https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines

joel-bluedata:
I talked to Swami a bit more about that... perhaps a good solution for such daemonsets would be to have them run as non-root too, and specify a limited set of sudo commands that they are allowed to do. That would limit concerns about root breakout from the container. I know you're thinking about this in the context of the fork you're working on, and I'd say that's a subsequent-phase (but before final shipping) sort of concern rather than a showstopper for initial implementation.

don-milos:
Since you said that we don't need to address the FsMount issue in the initial implementation can I assume these non-root user changes won't be merged into kubedirector-plus until the issue is addressed?

The reason I ask is that I don't know for sure but I would be surprised if these changes don't prevent the kubedirector user from deploying a privileged daemonset as it is currently implemented.

joel-bluedata:
Let's talk in email/skype.

PR #205, opened 2019-10-24
misc cleanup/changes from other forks

joel-bluedata:
Omnibus for some accumulated changes that are more trouble than they're worth to split up into separate PRs.

Change per file described below. If a modified file is not listed here, its change consisted only of an update to the copyright statement, indentation, comments, and/or various IDEs fooling around with how they want to assign import aliases.

#### Makefile

Enforce golang version 1.12 or later... necessary to compile successfully given some language constructs used.

Fix various indentations.

When checking for an existing KD during "make deploy", check all namespaces.

When deleting objects, check all namespaces.

Make various progress messages less verbose or more useful.

Allow the result of "build compile" to be used for redeploy even if done on non-Linux -- explicitly set GOOS to linux.

#### kubedirector_v1alpha1_kubedirectorcluster_crd.yaml

Add ClusterIP as a valid service type.

Add support for an optional secret to be mounted into the pods of one or more roles. The secret must exist in the cluster CR's namespace.

#### kubedirector_v1alpha1_kubedirectorconfig_crd.yaml

Add a status subresource. We expect more complex config reconciliation in the future; also this is a step toward unified status subresource handling (as far as write protection is concerned) for all monitored CRs.

Add ClusterIP as a valid default service type.

#### rbac-default.yaml

Add "watch" verbs to a couple of resources to suppress warnings from the new operator SDK framework.

#### kubedirector-development.md

Mention the requirement for minimum golang version 1.12.

#### kubedirectorcluster_types.go

Add support for the KDCluster secret described above.

#### kubedirectorconfig_types.go

Add the Status subresource.

Correctly mark Spec as optional (pointer type).

#### controller/kubedirectorcluster/cluster.go

Use the genericized finalizer-handling functions.

#### controller/kubedirectorcluster/kubedirectorcluster_controller.go

Reduce log spew.

#### controller/kubedirectorcluster/members.go

Fix bug in log message.

#### controller/kubedirectorconfig/config.go

New top-level reconciliation logic for the config CR, and management of the status subresource. Nothing complicated happening here yet, but setting the stage for future development of features that rely on periodic reconciliation that looks at the config CR.

#### controller/kubedirectorconfig/kubedirectorconfig_controller.go

Enable 30 second polling. Not useful yet (until more complicated config reconciliation), but pretty harmless given there will always only be at most one of these CRs.

Reduce log spew.

Call the new top-level reconciliation logic mentioned above.

#### executor/cluster.go

Remove cluster-CR-specific finalizer-handling functions in favor of new genericized versions.

#### executor/guest.go

Added a new RemoveDir utility function. The work that was going to make use of this hasn't been merged yet, but this is tested and works, so, leaving it in here.

#### executor/service.go

Changes to pod service spec to fix reverse DNS lookups of pod addresses on newer K8s versions. Harmless change for older K8s versions.

Changes to UpdatePodService to correctly handle some service-type transitions that require deleting and re-creating the service.

#### executor/statefulset.go

Add support for the KDCluster secret described above.

#### shared/kubedirectorobject.go

Generic finalizer-handling functions.

#### shared/service.go

String-to-type conversion factored out from the executor code.

#### shared/types.go

Finalizer-related constant moved to shared/kubedirectorobject.go with the generic finalizer-handling code.

Default service type constant moved here from the validator code.

New event reason for use during config CR reconciliation.

#### validator/cluster.go

Add support for the KDCluster secret described above.

#### validator/kdsettings.go

Handle the case where spec is entirely omitted.

Handle the status subresource.

Some refactoring of the patch generation.

#### validator/types.go

Support logging during validation.

Default service type constant moved out of here into the shared code.

Constants for KDCluster secret validation.

#### example_configs

Renamed this directory to be consistent with the other such directory names.


Makefile
```diff
@@ -63,34 +74,37 @@ push:
...
 
 deploy:
 	@set -e; \
-        pods_gone=False; \
-        kubectl get -o jsonpath='{.items[0].metadata.name}' pods -l name=${project_name} &> /dev/null || pods_gone=True; \
-        if [[ "$$pods_gone" != "True" ]]; then \
-            echo "KubeDirector pod already exists. Maybe old pod is still terminating?"; \
-            exit 1; \
-        fi; \
+        all_namespaces=`kubectl get ns --no-headers| awk '{print $$1}'`; \
```
don-milos: I assume kubedirector will only be running in one namespace? If so, you don't need to loop through all the namespaces. Even if there can be multiple pods, you can get a list of all of them and their namespaces with a single command:

kubectl get pods -l name=kubedirector --all-namespaces --no-headers | awk '{print $1, $2}'

joel-bluedata: About why we are checking everywhere: it's true that KD will only be deployed into one namespace, but there's no guarantee that the current active namespace where THIS deployment attempt is happening is the same as whatever namespace was used for the previous deployment attempt. Once some of us started to deploy KD into kube-system, this bit us a couple of times when changing over to kube-system after doing a test deployment in some other namespace.

joel-bluedata: Good point about the simpler get format. I think I will leave this PR as a straight copy of the changeset though, and do any improvements in a subsequent PR.

Makefile
```diff
@@ -174,41 +188,82 @@ redeploy:
...
+            delete_all_things $$ns ${cluster_resource_name}; \
+        done; \
+        echo; \
+        echo \* Deleting any application types...; \
+        for ns in $$all_namespaces; do \
+            delete_all_things $$ns ${app_resource_name}; \
+        done; \
+        echo; \
+        echo \* Deleting any configs...; \
+        for ns in $$all_namespaces; do \
```
don-milos: Same comment as above. It's unnecessary to loop through the namespaces. You call delete all of these resources with a single command.

kubectl delete kubedirectorconfig --all --all-namespaces --now

This obviously also applies to all of the other delete loops.

pkg/controller/kubedirectorconfig/kubedirectorconfig_controller.go
```diff
@@ -74,8 +76,8 @@ var _ reconcile.Reconciler = &ReconcileKubeDirectorConfig{}
 
 const (
 	// Period between the time when the controller requeues a request and
-	// when it's scheduled again for reconciliation. Zero means don't poll.
-	reconcilePeriod = 0
+	// it's scheduled again for reconciliation. Zero means don't poll.
+	reconcilePeriod = 30 * time.Second
```
don-milos: Without FsMount is there a reason to poll?

pkg/executor/statefulset.go
```diff
@@ -16,17 +16,18 @@ package executor
...
 	"k8s.io/apimachinery/pkg/types"
-	"path/filepath"
-	"strings"
 
 	kdv1 "github.com/bluek8s/kubedirector/pkg/apis/kubedirector.bluedata.io/v1alpha1"
 	"github.com/bluek8s/kubedirector/pkg/catalog"
 	"github.com/bluek8s/kubedirector/pkg/shared"
 	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
don-milos: Can we make the tag corev1?

pkg/executor/statefulset.go
```diff
@@ -429,7 +461,6 @@ func generateVolumeMounts(
 	nativeSystemdSupport bool,
 	persistDirs []string,
 ) ([]v1.VolumeMount, []v1.Volume, error) {
-
```
don-milos: I know you like to have a blank line at the beginning of each function. Did you intend to delete this one?

pkg/executor/volume.go
```diff
@@ -16,9 +16,10 @@ package executor
 
 import (
 	"context"
+
 	"github.com/bluek8s/kubedirector/pkg/shared"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
don-milos: Can you make the tag corev1?

Makefile
joel-bluedata: Tracking in issue #206.

Makefile
joel-bluedata: Tracking in issue #206.

pkg/executor/statefulset.go
joel-bluedata: Just bringing over the changes verbatim at the moment, unless we find functional problems.

I think this particular thing is something we need a more systematic solution to (if we want to enforce it), because several of us use IDEs that insert these import statements, and the IDEs are inclined to just alias this as "v1".

pkg/controller/kubedirectorconfig/kubedirectorconfig_controller.go
joel-bluedata: Nah, not yet. As I mentioned in the PR description this should be harmless, but not useful yet.

I've waffled on whether or not to disable the polling here; I don't want to later merge this change into some other fork that requires the polling and break it.

Maybe if we're both aware of that issue we can make sure it won't happen!

Issue #207, opened 2019-11-04
status update race observed?


joel-bluedata:
(Issue would be high priority if it could be reproduced...)

When testing on EKS I observed the following kind of behavior where duplicate per-role statefulsets are created from a single KDCluster creation:
```
{"level":"info","ts":1572893589.6170433,"logger":"controller_kubedirectorcluster","msg":"new","Request.Namespace":"default","Request.Name":"spark-persistent"}

{"level":"info","ts":1572893589.6330311,"logger":"controller_kubedirectorcluster","msg":"creating role{controller}","Request.Namespace":"default","Request.Name":"spark-persistent"}
{"level":"info","ts":1572893589.6421049,"logger":"controller_kubedirectorcluster","msg":"creating role{worker}","Request.Namespace":"default","Request.Name":"spark-persistent"}
{"level":"info","ts":1572893589.6496427,"logger":"controller_kubedirectorcluster","msg":"creating role{jupyter}","Request.Namespace":"default","Request.Name":"spark-persistent"}
{"level":"info","ts":1572893589.760274,"logger":"controller_kubedirectorcluster","msg":"changing replicas count for role{controller}: 0 -> 1","Request.Namespace":"default","Request.Name":"spark-persistent"}
{"level":"info","ts":1572893589.7761517,"logger":"controller_kubedirectorcluster","msg":"changing replicas count for role{worker}: 0 -> 2","Request.Namespace":"default","Request.Name":"spark-persistent"}
{"level":"info","ts":1572893589.793077,"logger":"controller_kubedirectorcluster","msg":"changing replicas count for role{jupyter}: 0 -> 1","Request.Namespace":"default","Request.Name":"spark-persistent"}

{"level":"info","ts":1572893589.8285778,"logger":"controller_kubedirectorcluster","msg":"creating role{controller}","Request.Namespace":"default","Request.Name":"spark-persistent"}
{"level":"info","ts":1572893589.836742,"logger":"controller_kubedirectorcluster","msg":"creating role{worker}","Request.Namespace":"default","Request.Name":"spark-persistent"}
{"level":"info","ts":1572893589.8455975,"logger":"controller_kubedirectorcluster","msg":"creating role{jupyter}","Request.Namespace":"default","Request.Name":"spark-persistent"}
[...]

PR #208, opened 2019-11-05
reinstate "stale CR protection"

joel-bluedata:
addresses issue #207

In the previous form of the operator framework, where the framework would deliver the CR contents to the reconcile handler as an argument, we had to check the incoming status UID to protect against the following case: a reconcile invocation (usually just from a periodic poll) was queued up right as we were making changes to the CR in a current handler, so the CR delivered to the next handler would then not contain the changes we made.

In the new operator framework we didn't think we needed this. Each handler fetches the CR when it starts running, through the controller-runtime client. That client does some caching but we believed that it was guaranteed that when we do a PUT in one handler followed by a GET in a later handler, the GET would show the results of the PUT.

Unfortunately it doesn't seem like we can depend on this, unless there is some other bug that explains the behavior I'm seeing. This is something I have observed occasionally when running KD on EKS (K8s version 1.14):
- Some client creates a KubeDirectorCluster.
- KD runs a handler to deal with the create. It writes a status UID to the KDCluster CR.
- This write triggers another KD handler to run for that same CR. However, when it does a GET of the CR it doesn't see any status UID.

I've also seen a similar case where a handler creates the statefulset roles and records them in the status, and then the next handler doesn't see those roles in the status stanza so it creates another set of roles.

This is rare but not SUPER rare in my EKS experiments. I've never seen it when running in local K8s clusters or in GKE, but I don't know if that is meaningful.

I can't help but wonder if something weird is going on in the caching client w.r.t. how the status subresource is treated. We are writing to the status subresource but then reading the whole CR later. I don't completely understand the issue, but I do know that the "stale CR protection" keeps it from affecting us. It's definitely not a matter of our status writes being lost entirely, because when I put in the stale CR protection to abort that handler and try the read again, everything works. So I propose we put that protection back in.

The only difference between this implementation and the previous one is that I am returning an error from the reconcile handler so that it will be *immediately* retried, rather than waiting for the next poll.

Also I'm adding this to the config CR processing as well as the cluster CR, since the config CR now uses status UID. Seems like it's a common mechanism we want to use in all such cases.

### DIFF
--- a/pkg/controller/kubedirectorcluster/cluster.go
+++ b/pkg/controller/kubedirectorcluster/cluster.go
@@ -15,6 +15,7 @@
 package kubedirectorcluster
 
 import (
+	"fmt"
 	"reflect"
 	"time"
 
@@ -114,7 +115,10 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 	}()
 
 	// For a new CR just update the status state/gen.
-	shouldProcessCR := r.handleNewCluster(reqLogger, cr)
+	shouldProcessCR, processErr := r.handleNewCluster(reqLogger, cr)
+	if processErr != nil {
+		return processErr
+	}
 	if !shouldProcessCR {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


joel-bluedata:
Yeah for forks that have other kinds of CRs we'd want to do the same thing there too. Ideally all this status UID handling can be factored into a shared subroutine.

joel-bluedata:
I'm going to take a quick whack at generalizing the code.

PR #211, opened 2019-11-05
camelCase our CR property names

joel-bluedata:
This follows the style guide for Kubernetes resource properties, and "kubectl describe" works best with camelCased names. So let's bite the bullet.

One bullet not bitten here however is the naming scheme for the config data seen inside the container by config CLI. Changing those property names, although possibly nice for consistency with some parts of the CRs, would require also changing config CLI and app startscripts. I'm not sure if we want to do that, and even if we do, let's not bundle it into this change here.

### DIFF
--- a/pkg/apis/kubedirector.bluedata.io/v1alpha1/kubedirectorapp_types.go
+++ b/pkg/apis/kubedirector.bluedata.io/v1alpha1/kubedirectorapp_types.go
@@ -15,7 +15,7 @@
 package v1alpha1
 
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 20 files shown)


pkg/apis/kubedirector.bluedata.io/v1alpha1/kubedirectorapp_types.go
```diff
@@ -101,26 +101,26 @@ type ServiceEndpoint struct {
...
 
 // NodeGroupConfig identifies a set of roles, and the services on those roles.
 // The top-level config indicates which roles and services will always be
 // active. Implementation of "config choices" will introduce other conditional
 // configs.
 type NodeGroupConfig struct {
-	RoleServices   []RoleService     `json:"role_services"`
-	SelectedRoles  []string          `json:"selected_roles"`
-	ConfigMetadata map[string]string `json:"config_meta"`
+	RoleServices   []RoleService     `json:"roleServices"`
```
distortedsignal: Do you think it would be possible to add this as a check in the Travis job? Would that be a good idea?

distortedsignal:
I'm down with this change, even if we don't add it to the Travis job.

PR #212, opened 2019-11-06
Delete operator-sdk generated files from git.

don-milos:


.travis.yml
```diff
@@ -5,6 +5,7 @@ go:
 
 env:
   - DEP_VERSION="0.5.2"
+  - OPERATOR_SDK_VERSION="v0.8.1"
```
distortedsignal: It looks like the `env` describes one-build-per-line. So this might have to be like this instead:

```yaml
env:
  - DEP_VERSION="0.5.2" OPERATOR_SDK_VERSION="v0.8.1"
```

That's an... odd... choice by Travis, but I understand why it's necessary. I wish there was a better way to check that this sort of thing would happen _before_ pushing.

Makefile
```diff
@@ -312,7 +312,7 @@ compile: pkg/apis/kubedirector.bluedata.io/v1alpha1/zz_generated.deepcopy.go ver
 	GOOS=linux GOARCH=${goarch} CGO_ENABLED=${cgo_enabled} \
         go build -o ${build_dir}/bin/${bin_name} ./cmd/manager
 
-ci-compile:
+ci-compile: pkg/apis/kubedirector.bluedata.io/v1alpha1/zz_generated.deepcopy.go
```
joel-bluedata: At this point should we just get rid of the `ci-compile` target and just use the `compile` target in Travis too?

joel-bluedata: Although (seeing your latest commit) I'm curious about why a make clean was needed on Travis.

distortedsignal: Looking at the commands, the differences are _somewhat minimal_:

```Makefile
compile: pkg/apis/kubedirector.bluedata.io/v1alpha1/zz_generated.deepcopy.go version-check
    -rm -f ${build_dir}/bin/${bin_name}
    GOOS=linux GOARCH=${goarch} CGO_ENABLED=${cgo_enabled} \
        go build -o ${build_dir}/bin/${bin_name} ./cmd/manager


ci-compile: clean pkg/apis/kubedirector.bluedata.io/v1alpha1/zz_generated.deepcopy.go
    GOOS=linux GOARCH=${goarch} CGO_ENABLED=${cgo_enabled} \
        go build -o ${build_dir}/bin/${bin_name} ./cmd/manager
```

Clean in Travis should do _almost nothing_, but I like having it around for safety. I'm curious what the `version-check` tool does and if it's installed in Travis by default. I think this change should be kept as small as possible, and since we've got a working build, we should _probably_ merge this now. Some exploratory work could be done to change `ci-compile` to be something like

```Makefile
ci-compile: clean compile
```

Or remove the `ci-compile` target altogether and change the `.travis.yml` file to do `make clean compile`. I think I like keeping the `ci-compile` target, because it lets people run the same command as Travis will in one simple step. Interested to hear your thoughts.

don-milos: version-check just validates the version of the compiler being used. We should do that in travis and non-travis compiles so that difference goes away.

The only difference after that is:

1. - The ci-compile target currently does a clean. If you look at the clean target it removes rbac.yaml, deployment-localbuilt.yaml, (I just added this) zz_generated.deepcopy.go, and the build directory.
2. - The compile target just removes the kubedirector binary

I don't think we want to remove the yaml's and zz_generated.deepcopy.go source file for a normal compile and I don't think the yaml's will even exist in the travis build. So the only real affect of doing a clean for a ci-build is that it removes the entire build directory.

I think we should combine the 2 targets and have the combined compile target remove the entire build directory instead of doing a clean That won't really change the behavior of the current compile target or the current ci-compile target. 

Does that make sense or did I miss something?

don-milos: I'm going to try that change to see how the build goes. I can back it out if you guys don't like it.

don-milos: Fyi, the build worked after combining the compile targets.

joel-bluedata: Makes sense to me!

PR #218, opened 2019-11-14
fixes for CI

joel-bluedata:
Primarily this is to get CI working on forks again (after the generated deepcopy code was removed from the repo), although also some general-goodness changes included.

* Actually move the cloned repo to the desired directory (bluek8s/kubedirector) rather than just making a symlink. At one point I thought this was needed to fix some operator-sdk behavior, now I'm not sure, but things work so I'm going to leave it this way.

* Don't do "make dep" (which does "dep ensure -v -update"); just do "dep ensure -v".

* Move some stuff from before_install into the install step. Partly just because these things make sense there, but also because if we don't define the install step, the default install step (which runs travis_install_go_dependencies) tries to do things that depend on the generated deepcopy code, before it is generated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,20 +4,23 @@ go:
   - "1.12"
 
 env:
-  - DEP_VERSION="0.5.2" OPERATOR_SDK_VERSION="v0.8.1"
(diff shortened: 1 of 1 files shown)


distortedsignal:
Well, it looks like this make the build pass on your fork, so I'm for it.

PR #233, opened 2019-12-05
KubedirectorCluster/Model Attachments

kmathur2:
What are attachments ?

Currently there are 2 types of attachments -
- Cluster attachments
- Model attachments

**Cluster attachments** - When we create a kubedirectorcluster we generate "configmeta" as part of cluster creation which has all cluster related metadata. There are tools like configcli and bdvcli to query this metadata from within a kubedirectorcluster pod. As part of this feature if some other cluster pod wants to query metadata from another cluster we use this attachment feature. As part of the feature kubedirectorapp can define the allowed attachment condition (example cluster category, distro_id etc) based on which a given running cluster could be attached to the current cluster. As part of creating a new kubedirectorcluster the user can specify a list of kubedirectorcluster to be attached in this cluster's spec. The kd cluster reconciler will then dump the "cluster to be attached"'s configmeta in the new cluster and it could be then queried using conflicli from within the current cluster's pod.   

**Model attachments** - As part of current cluster spec a list of data science model metadata (model name, model version, model location on the tenant mounted file system, path to model scoring script on the tenant mounted file system) could also be attached. Again this model metatdata could be queried from with current cluster's pod using configcli. In future this model metatdata might be its own custom resource OR a config map.


joel-bluedata:
This feature can lag a bit behind some of the other stuff we're on deadline for, so it isn't super high priority for review yet FYI. One thing I do want to mention is that we can slow-walk it juuust a bit to make sure we get the model right. The model here is lifting some stuff directly from EPIC and/or our operator hackathon and might benefit from another look.

For example the way that there are two explicit lists of cluster and model attachments in the cluster CR ... is that the way we should model this in K8s-land? Something we should chat about in person when we have free time. Also of course we'll need to update the CRDs here when we are sure of the model.

kmathur2:
- [ ] Fix domain Base

- [ ] Clean up reconciller code

- [ ]

kmathur2:
Closing this in lieu of https://github.com/bluek8s/kubedirector/pull/283

PR #235, opened 2019-12-12
Support default KD apps for Airgap clusters 

kmathur2:
Currently the default KD Apps pull appconfig from s3://. However it will not be possible to pull appconfig for air-gap k8s cluster. Hence for the default KD Apps the appconfig will part of the image and not a separate entity which needs to be pulled at cluster create time. 
This PR has changes for cassandra and spark221. There will be a separate PR for cdh (since india team is already working on dtap:// support for cdh) and latest tensorflow. 

- Also removed current  tensorflow cpu and gpu apps in lieu of new tensorflow cpu/gpu app with other updated datascience libraries (Separate PR for this later)


.gitignore
```diff
@@ -1,5 +1,6 @@
 /Local.mk
 
+*.tgz
```
joel-bluedata: Is this specifically for build/configcli.tgz? Maybe we should be more specific.

joel-bluedata:
Also... do we need to remove those tensorflow apps right now, or can we wait until we have the new ones ready?

.gitignore
kmathur2: Yes it is, do we want to put any tgz ?

kmathur2:
> Also... do we need to remove those tensorflow apps right now, or can we wait until we have the new ones ready?

We will have the other one pretty soon, may be today or tomorrow. Also this TF is old and not well tested

.gitignore
joel-bluedata: Dunno, I'm a little skittish about really broad ignore rules because they can result in inadvertent failures to check in files in the future.

kmathur2: ok sure

PR #242, opened 2020-01-17
user-specified labels feature tweaks

joel-bluedata:
Add serviceLabels to the role config too, to allow specifying labels for services related to the role.

Also a couple of slight tweaks to existing label behavior:

- Don't put the user-specified pod member labels on the statefulset; only on the pods.
- Go easier on the labels in the pod selector in a member service; just reference the pod name.

### DIFF
--- a/pkg/apis/kubedirector.bluedata.io/v1alpha1/kubedirectorcluster_types.go
+++ b/pkg/apis/kubedirector.bluedata.io/v1alpha1/kubedirectorcluster_types.go
@@ -104,7 +104,7 @@ type FileInjections struct {
 // defined by the cluster's KubeDirectorApp) set of service endpoints.
 type Role struct {
 	Name           string                      `json:"id"`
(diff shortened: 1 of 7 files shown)


pkg/apis/kubedirector.bluedata.io/v1alpha1/kubedirectorcluster_types.go
```diff
@@ -105,6 +105,7 @@ type FileInjections struct {
 type Role struct {
 	Name           string                      `json:"id"`
 	Labels         map[string]string           `json:"labels,omitempty"`
```
swamibluedata: Should we call this PodLabels now?

joel-bluedata: Maybe! Yeah, I did ponder that. It does make sense.

Issue #253, opened 2020-01-30
files on persisted dirs can be overwritten on container restart


joel-bluedata:
We've got an issue with the change from PR #243: when it creates the marker file kubedirector.init, it creates it in the root filesystem of the init container rather than in the mounted dir from the PV. So next time around when the container is restarted, the marker file will be gone. This means the copy-from-image will happen again, so any file that originally existed on the image will be reverted to its original state. (Newly created files are ok.)

The fix should be simple, just changing the location of the marker file in generateInitContainerLaunch.

PR #254, opened 2020-01-30
persistence fix

joel-bluedata:
For issue #253 .

### DIFF
--- a/deploy/example_catalog/cr-app-centos7.json
+++ b/deploy/example_catalog/cr-app-centos7.json
@@ -7,6 +7,7 @@
 
     "spec" : {
         "systemdRequired": true,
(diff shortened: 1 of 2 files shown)


deploy/example_catalog/cr-app-centos7.json
```diff
@@ -7,6 +7,7 @@
 
     "spec" : {
         "systemdRequired": true,
+        "defaultPersistDirs" : ["/home"],
```
swamibluedata: just for reference?

joel-bluedata: Just to give it more interesting behavior when using persistent storage. (Should we add some other dirs too?)

PR #262, opened 2020-02-03
API group, labels, and terminology cleanup.

joel-bluedata:
* Moving to an API group in a domain with better longterm security (based on hpe.com) and also promoting to v1beta1.

* Adding "shortname" alternatives for the CRs (kdcluster, kdapp, kdconfig).

* Making object naming more consistent; cf. https://github.com/bluek8s/kubedirector/commit/04544b30de99821373af44b79426da063e9472c0

* Improvements/changes/additions for labelling of created objects: https://github.com/bluek8s/kubedirector/commit/0e6a639b54ccc96f6fb1994d102c28006d3b907c

* Update deps.


Makefile
```diff
@@ -45,7 +45,7 @@ version-check:
         exit 1; \
     fi
 
-build: configcli pkg/apis/kubedirector.bluedata.io/v1alpha1/zz_generated.deepcopy.go version-check | $(build_dir)
+build: configcli pkg/apis/kubedirector.hpe.com/v1beta1/zz_generated.deepcopy.go version-check | $(build_dir)
```
swamibluedata: why not v1?

pkg/executor/types.go
```diff
@@ -23,24 +23,42 @@ import (
...
-	kubedirectorInit    = "/etc/kubedirector.init"
+	ClusterRoleLabel = shared.KdDomainBase + "/role"
+	// HeadlessServiceLabel is a label placed on the statefulset and pods.
+	// Used in a selector on the headless service.
+	HeadlessServiceLabel = shared.KdDomainBase + "/headless"
+
+	// ClusterAppAnnotation is an annotation placed on every created
+	// statefulset, pod, and service, with a value of the KubeDirectorApp's
+	// spec.label.name.
+	ClusterAppAnnotation = shared.KdDomainBase + "/kdapp-prettyName"
```
swamibluedata: ?

joel-bluedata: spec.label.name from the kdapp. Can't store it as a label value because it is an arbitrary string.

Makefile
joel-bluedata: Conservative. :-) I'd like this to soak for a while with the "beta" tag.

We also have quite a few API-affecting features in the wings, and while we should be able to do this in a backwards-compatible manner, if we're really going to churn the API for a while yet then I'd rather bump the version v1beta1->v1beta2->v1beta3 etc. rather than v1->v2->v3.

PR #264, opened 2020-02-05
Fix to enable GPU usage in Jupyter with TensorFlow

saurabh-jogalekar:
Summary of changes:

- The app now uses official `tensorflow-jupyter-gpu` docker image. [Link ](https://hub.docker.com/r/tensorflow/tensorflow/tags?page=1&name=latest-gpu-py3-jupyter)
- Removed role `worker` since it is not required
- GPU is a required resource. 
- TensorFlow version is now dependent on the version of the base image 

Following steps are required to start the notebook image:

1. Create the kubedirectorapp and kubedirectorcluster
2. Run the command `kubectl exec <pod-name> jupyter notebook list` to get the jupyter notebook authentication token.
3. Run jupyter notebook


deploy/example_clusters/cr-cluster-tensorflow-notebook.yaml
```diff
@@ -10,15 +10,12 @@ spec:
...
-      requests:
-        memory: "4Gi"
-        cpu: "2"
-      limits:
-        memory: "4Gi"
-        cpu: "2"
+        nvidia.com/gpu: "1"        
+    env:
+      - name : "LD_LIBRARY_PATH"
+        value: "/usr/local/nvidia/lib64"
```
joel-bluedata: Hmm. If the user is required to enter this in the kdcluster spec each time, it's too brittle for usability IMO. Does the official image really not work without this?

Some other solutions:

- Modifying the image to include this in /etc/environment or some other relevant profile. Obviously this is a bummer if you want to use the unmodified image.
- Adding a startscript package that does that modification at runtime. Should work fine but does require the startscript package residing on some available http server. A bummer if we want our example kdapps to be self-contained.
- New feature in KD to allow specifying env vars in the role definition of the KubeDirectorApp. Final set provided to the pod would be the env vars from the kdapp unified with (overridden by) any env vars from the kdcluster.

Let me know your thoughts.

kmathur2: #2 is doable by having start-script baked into the image. Although #3 is ideal, might not be possible for 5.0 if we want this to be shipped

kmathur2: hmm, actually it will 1st not 2nd then.

joel-bluedata: First we should determine if it really is required, i.e. if the official image is broken without specifying this. I would be surprised-ish, but stranger things have happened.

swamibluedata: This is what i had before

https://github.com/bluek8s/kubedirector/blob/v0.3.0/deploy/example_catalog/cr-app-tensorflowgpu.json

This just worked without any changes

saurabh-jogalekar: Yes, it works even without the environment variable. I'll remove it.

joel-bluedata:
One more thought... could add an annotation to the app CR that describes the necessary step to get the jupyter notebook authentication token.

Other than that, looks good! I'll merge once you've synced this back up to the tip of master branch.

PR #281, opened 2020-03-02
Datatap support for Spark KD app

madhusudhannr:
To support dtap in spark kubedirector app - kd-Spark221-appconfig, following changes have been made.

1. updated start script file to create /opt/bdfs/ folder and copy related jar files to the created folder.
2. updated java, spark and dtap paths in hadoop, macros.sh and spark-defaults.conf files to above created folder.
3. .tgz files of sparkbase and Jupyter has been udpated and uploaded to Docker repository with versions changed from bluedata/sparkbase:2.2 to bluedata/sparkbase:2.3 and from bluedata/jupyter:2.3 to bluedata/jupyter:2.4 respectively.
4. Updated related app json file with updated version for sparkbase and Jupyter (cr-app-spark221e2.json in Everest)

### DIFF
--- a/deploy/example_catalog/cr-app-spark221e2.json
+++ b/deploy/example_catalog/cr-app-spark221e2.json
@@ -7,7 +7,6 @@
 
     "spec" : {
         "systemdRequired": true,
-        "defaultPersistDirs" : ["/usr", "/opt", "/var", "/data"],
         "config": {
             "roleServices": [
                 {
--- a/deploy/example_catalog/cr-app-spark221e2.json
+++ b/deploy/example_catalog/cr-app-spark221e2.json
@@ -106,7 +106,7 @@
                 }
             }
         ],
-        "defaultImageRepoTag": "docker.io/bluedata/sparkbase:2.2",
+        "defaultImageRepoTag": "docker.io/bluedata/sparkbase:2.3",
         "defaultConfigPackage":  {
             "packageURL": "file:///opt/configscripts/appconfig-2.6.tgz"
         },
@@ -120,7 +120,7 @@
                 "id": "worker"
             },
             {
-                "imageRepoTag": "docker.io/bluedata/jupyter:2.3",
+                "imageRepoTag": "docker.io/bluedata/jupyter:2.4",
                 "cardinality": "0+",
                 "id": "jupyter"
             }


joel-bluedata:
There's another thing I'm interested in improving in our example Spark app; maybe we should tackle that here, or maybe in another PR. Let me mention it FYI:

Currently the persistDirs are defined only at the top level spec (in defaultPersistDirs). This means that the app spec is declaring that the controller, worker, and jupyter roles all have the same set of directories that they need to persist. So:
a) I am really dubious that this is accurate. Do the workers even need persisted directories at all? Does Jupyter need a different set of directories.
b) The current set of directories (and the data in them at startup) is quite large, leading to multi-minute initial startup times. Can we narrow this down? E.g. do we really need all of /usr and /var or could we specify subdirectories? (We also need to narrow down what KD itself mounts to support bdvcli, but we don't want the app to become the bottleneck when we do that.)

kmathur2:
I think we should have a way to manage source control for guest-config scripts as well. For some KD apps we have them in bluedata-catalog repo but its still adhoc. Shall we continue to manage them in that repo ? If yes then we need to standardize a way for referencing guestscripts for a given app to bluedata-catalog repo and open PRs accordingly. Any thoughts/suggestions @joel-bluedata

riteshja:
Hi @joel-bluedata 

We have removed `persistDirs` from the top level spec altogether. So now we are not persisting any directories and tested the app with reboot scenarios and running jobs. Both notebook and spark are working fine. Is this okay?

joel-bluedata:
You don't have issues with the master node when it is restarted and loses all of its data?

joel-bluedata:
Specifically with any ongoing jobs?

joel-bluedata:
It also sounds bad to not have persistence for the notebook role at all. There will be directories where user data gets stashed in there.

riteshja:
Quick question-

Does KD persist any directories by default for the pods in a KDCluster?

joel-bluedata:
Not in way that app authors could/should depend on.

riteshja:
Alright. We are doing a major refactor of this app. :) And also updating the spark version to 2.4.5.
We will be back soon. Till then, closing this PR.

madhusudhannr:
We are doing a major refactor of this app. And also updating the spark version to 2.4.5.
We will be back soon. Till then, closing this PR.

Issue #265, opened 2020-02-07
bad interaction between pod eviction and leader lock


joel-bluedata:
https://github.com/operator-framework/operator-sdk/issues/1305
https://github.com/operator-framework/operator-sdk/issues/1874

We've hit this.

Fixed by https://github.com/operator-framework/operator-sdk/pull/2210

We won't get this until operator SDK v0.13. Until then I think we will just have to comment out the invocation of the leader mechanism in main.go.

joel-bluedata:
Alternatively, the patch is simple enough that perhaps we could backport it into a fork of "our" version of the SDK.

But I'm not sure how much it buys us. The leader lock itself is just operating via K8s mechanisms and state (a configmap), so I don't think it helps in the case where there's a single deployment. It does protect against multiple deployments, which is nice but maybe not super compelling.

joel-bluedata:
Note that they've also previously had deadlock between leader election and a readiness probe: https://github.com/operator-framework/operator-sdk/issues/920

We propose a liveness probe and not a readiness probe, so that could be fine, but certainly something to test (if we're not going to just comment out the leader election).

joel-bluedata:
For the 0.4.0 release we are just not doing leader election.

Removing the milestone on this one for now. We'll naturally get a fix if/when we update the SDK to v0.13, but maybe we'll also be compelled to try some fix for it sooner.

joel-bluedata:
This is still problematic even with the new operator code. Pod is not ever getting evicted when the node is turned off, undoubtedly because my test cluster is too small (less than large-cluster-size-threshold).

So hmm. I'll experiment with the leader-with-lease approach.

joel-bluedata:
Leader-with-lease looks good. Will close this once #282 is merged.

PR #282, opened 2020-03-10
use latest Operator SDK (0.15.2)

joel-bluedata:
Putting this PR here to air out a bit while we test it more.

Lots of go source files got touched in the "imports" section because of the directory name change from "pkg/apis/kubedirector.hpe.com" to "pkg/apis/kubedirector". Other changes listed below.

#### .gitignore

Removing "dep"-related stuff.

Some formatting changes in the autogenerated part.

#### .travis.yml

New golang version requirement.

Switching from dep to go modules.

#### Gopkg.{lock,toml} and go.{mod,sum}

Switching from dep to go modules.

#### Makefile

Updating the golang version requirement & adding an operator-sdk version requirement.

Removing "dep"-related stuff.

Directory name change pkg/apis/kubedirector.hpe.com -> pkg/apis/kubedirector.

Make clean should also remove configcli.tgz.

Fixes for .PHONY.

#### build/*

Slightly different skeleton generated by current SDK -- user setup is done at image build time. Not 100% sure this will still be fine under OpenShift strictness but it is what the SDK now recommends.

#### doc/*

Updating docs for changes to supported/suggested K8s platforms and changes in build process.

Adding upgrade.md to contain notes about moving to the new version.

#### main.go

Re-enabling leader election, but "with lease" rather than "for life"; cf. issue #265.

New Prometheus metrics stuff. Commented out for now until we can actually test this.

#### deployment-prebuilt.yaml and operator.yaml

No predefined metrics port anymore.

WATCH_NAMESPACE really should be empty (since we're cluster scope), even though we don't use this.

#### rbac-default.yaml

Adding privs for services/finalizers since the new generated role includes that.

Also adding get/watch/list privs for daemonsets.

Adding privs for deployments/finalizers on our own deployment.

Adding Prometheus-related privs (monitoring.coreos.com).

#### cert.go

Had to pull in these functions we use that have been removed from latest client-go. We could/should move to using the tls module now provided by the SDK... however, its interface requires storing the various certs/keys in two secrets and a configmap, while our current method uses one secret. Since there are now some existing deployments of KD, we would need a migration strategy.

#### shared/client.go

Handling changes in the signatures for List and Delete.

Handling cases where the split client now can return ErrCacheNotStarted.

#### validator/util.go

For the Webhooks list element inside our MutatingWebhookConfiguration: instead of a Webhookobject we now are required to use MutatingWebhook. Even though the API version of MutatingWebhookConfiguration has not changed. This is a compatible change i.e. MutatingWebhook is a superset of the old Webhook.


joel-bluedata:
Note that this change does not take advantage of any new capabilities of the new SDK yet. Some things to investigate: using markup to help generate validation schema, secondary resources, CR status conditions, and the metrics stuff mentioned above.

PR #283, opened 2020-03-11
KubedirectorCluster Connections

kmathur2:
What are the Connections ?

Currently, there are 2 types of connections -

Cluster connections
Configmap connections

**Cluster connections** - When we create a kubedirectorcluster we generate "configmeta" as part of cluster creation which has all cluster related metadata. There are tools like configcli and bdvcli to query this metadata from within a kubedirectorcluster pod. As part of this feature if some other cluster pod wants to query metadata from another cluster we use this connection feature. As part of the feature, kubedirectorapp can define the allowed attachment condition (example cluster category, distro_id, etc) based on which a given running cluster could be connected to the current cluster. As part of creating a new kubedirectorcluster the user can specify a list of kubedirectorcluster to be connected to this cluster's spec. The kdcluster reconciler will then dump the "cluster to be attached"'s configmeta in the new cluster and it could be then queried using conflicli from within the current cluster's pod.

**Configmap connections** - As part of the current cluster spec a list of configmaps (model, external auth, Kerberos, etc) could also be connected. A connected configmap will be propagated inside the running pods of the parent kdcluster and could then be queried using bdvcli or similar tools. 

Example of connections at play, as part of kdcluster spec
```
spec:
  connections:
    configmaps:
    - "model-prediction"
    - "noise-recognition"
    clusters:
    - "spark-logs-nightly"
    - "kafka-streaming"
```

**Change Summary:**
In **KubeDirectorAppSpec** struct added a new field
   - ConnectableTo: This enumerates a list of allowed categories for connectable clusters

In **KubeDirectorClusterSpec** struct added 2 new fields
   - ConfigMetaGenerator: Everytime a new connection is added/deleted this is bumped up
   - Connections: New struct that carries a list of config maps and clusters to be connected

In **KubeDirectorClusterStatus** struct added 1 new field
   - LastConfigMetaGenerator: After every reconfig, this captures the last value of 
     ConfigMetaGenerator

**Added reconciler** for config maps to watch on any changes to the config map in the cluster namespace
**Every time connected configmap or cluster** is updated in the spec configMetaGenerator is incremented. 
**The cluster reconciler periodically looks for the value of configMetaGenerator** and compares against lastConfigMetaGenerator, if the value has changed then it triggers regeneration of configmeta and recreats configmeta.json file for all pods in the running cluster
**configMetaGenerator** will be incremented every time a connection is added/removed. 

**FollowUp tasks**  
- [ ] Need to add guesconfig hooks for configmap update
- [ ] Finalize on connectable_to property
- [ ] Reconciler refactoring using Enque fun
- [ ] Hash approach instead of configMeta incrementer  
- [ ] Settle if configmap should even be connection or can be mounted like secret or vice versa
- [ ] Finalize on v1Beta2 API


kmathur2:
yea, not this example. Need to do some cleanup and naming stuff wrt old naming

pkg/catalog/configmeta.go
```diff
@@ -116,6 +136,98 @@ func servicesForRole(
...
+	return kdcm, nil
+}
+
+// genClusterConnections generates a map of running clusters that are to be connected
+// to this cluster.
+func genClusterConnections(
+	cr *kdv1.KubeDirectorCluster,
+) (map[string]clusterConnections, error) {
+
+	// thisApp, _ := observer.GetApp(cr.Namespace, cr.Spec.AppID)
```
kmathur2: Question: Do we want the connectable_to kind of functionality as we did for EPIC? Personally I do not feel the need for it so commented for now

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
```diff
@@ -25,11 +25,20 @@ import (
...
+	DefaultSecret       *KDSecret   `json:"defaultSecret"`
+	ConfigMetaGenerator int32       `json:"configMetaGenerator"`
+	Connections         Connections `json:"connections"`
+}
+
+// Connections specifies list of cluster objects and configmaps objects that has
+// be connected to the cluster.
+type Connections struct {
+	Clusters   []string `json:"clusters,omitempty"`
+	ConfigMaps []string `json:"configmaps,omitempty"`
```
joel-bluedata: I think this is a case where generalizing-from-two is premature or incorrect. We handle these two kinds of "connections" somewhat differently, and there's no guarantee that there will be a third kind or how we might handle that. (I realize we sort of inherited this arrangement from EPIC, but I think we talked before about how it is probably not the right thing.)

I would recommend that we have these two lists NOT bundled together inside a Connections object. E.g. Connections (representing ONLY connected kdclusters) and ConfigMaps would be two separate and parallel properties with no wrapper object around them.

We should also have a serious think about whether either or both of these are the sort of thing that can be expressed per-role -- as is the case with secrets, labels, injected files, etc. etc. -- and not just at the top level spec.

Those decisions won't have much effect on the underlying logic of processing the configmaps and cluster connections, but we'll need to settle them before merging.

(Also of course there's the matter of making changes to the CRD. We have already put the stamp on v1beta1, so any changes we make here will need to be expressed in v1beta2. There's some stuff to figure out there in terms of how we author the CRD to express both v1beta1 and v1beta2, how we deal with existing resources, how we upgrade existing CRDs, and how we internally "promote" any v1beta1-styled resource inputs to store them as v1beta2. You and I and I think Koshik need to have a separate talk about that.)

pkg/apis/kubedirector/v1beta1/kubedirectorapp_types.go
```diff
@@ -27,12 +28,23 @@ type KubeDirectorAppSpec struct {
...
-	SystemdRequired     bool                `json:"systemdRequired,omitempty"`
+	DefaultPersistDirs  *[]string           `json:"defaultPersistDirs"`
+	Capabilities        []v1.Capability     `json:"capabilities"`
+	SystemdRequired     bool                `json:"systemdRequired"`
+	ConnectableTo       []ConnectableConfig `json:"connectable_to,omitempty"`
+}
+
+// ConnectableConfig describes type of objects that can be connected to
+// a cluster that uses the app.
+// XXX FIXME. Only categor
```
joel-bluedata: If this isn't used anywhere yet, let's remove it.

pkg/apis/kubedirector/v1beta1/kubedirectorapp_types.go
```diff
@@ -27,12 +28,23 @@ type KubeDirectorAppSpec struct {
...
-	Services            []Service           `json:"services,omitempty"`
+	Services            []Service           `json:"services"`
 	NodeRoles           []NodeRole          `json:"roles"`
 	Config              NodeGroupConfig     `json:"config"`
-	DefaultPersistDirs  *[]string           `json:"defaultPersistDirs,omitempty"`
-	Capabilities        []corev1.Capability `json:"capabilities,omitempty"`
-	SystemdRequired     bool                `json:"systemdRequired,omitempty"`
+	DefaultPersistDirs  *[]string           `json:"defaultPersistDirs"`
+	Capabilities        []v1.Capability     `json:"capabilities"`
+	SystemdRequired     bool                `json:"systemdRequired"`
```
joel-bluedata: Why the removal of all the omitempty tags?

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
```diff
@@ -25,11 +25,20 @@ import (
...
-	AppCatalog    *string   `json:"appCatalog,omitempty"`
-	ServiceType   *string   `json:"serviceType,omitempty"`
-	Roles         []Role    `json:"roles"`
-	DefaultSecret *KDSecret `json:"defaultSecret,omitempty"`
+	AppID               string      `json:"app"`
+	AppCatalog          *string     `json:"appCatalog"`
+	ServiceType         *string     `json:"serviceType"`
+	Roles               []Role      `json:"roles"`
+	DefaultSecret       *KDSecret   `json:"defaultSecret"`
+	ConfigMetaGenerator int32       `json:"configMetaGenerator"`
```
joel-bluedata: "configmeta" is kind of a confusing inside-baseball term for the JSON file, made a little more confusing by there also being a configMeta dictionary property inside the KD App.

IIRC this property, and the associated status property below, are intended to track changes specifically to the connected ConfigMaps. If that's the case then let's reflect that in the property name, e.g. we could call this one configMapGeneration and the status property could be lastConfigMapGeneration.

Also, can we store these as int64 rather than int32?

pkg/catalog/configmeta.go
joel-bluedata: Ah I see this is why we had that dangling type declaration before.

I think we do want declarations in the app about what kdclusters and configmaps could/should be connected to. But we should do a fresh design discussion on that (i.e. not copying the EPIC stuff necessarily) and tackle it in some subsequent PR. The feature works without that stuff, it's more of an additional feature.

pkg/catalog/configmeta.go
```diff
@@ -182,11 +297,21 @@ func clusterBaseConfig(
 	appCR *kdv1.KubeDirectorApp,
 	membersForRole map[string][]*kdv1.MemberStatus,
 	domain string,
-) *configmeta {
+) (*configmeta, error) {
+
+	clustersMeta, connErr := genClusterConnections(cr)
+	kdConfigMaps, cmErr := genconfigConnections(cr)
```
joel-bluedata: Similar to my comments above about the CR property... I think we can take this opportunity, if we want to, to improve the JSON representation of these things. I'm not familiar enough with that though to know what we might need to do.

pkg/catalog/configmeta.go
```diff
@@ -116,6 +136,101 @@ func servicesForRole(
...
 }
 
+// genconfigConnections will look at the cluster spec
+// and generates a map of configmap type and corresponding
+// configmaps to be connected to the given cluster
+func genconfigConnections(
+	cr *kdv1.KubeDirectorCluster,
+) (map[string]map[string]map[string]string, error) {
+
+	models := make(map[string]map[string]string)
```
joel-bluedata: Let's not use the term model in our var names, since these data structures are more generic than that.

pkg/catalog/configmeta.go
joel-bluedata: But e.g. we don't need to and probably shouldn't stick them in a wrapper "Connections" namespace together.

pkg/catalog/configmeta.go
```diff
@@ -116,6 +136,101 @@ func servicesForRole(
...
+	kdcm := make(map[string]map[string]map[string]string)
+	for _, modelMapName := range cr.Spec.Connections.ConfigMaps {
+		modelCM, err := observer.GetConfigMap(cr.Namespace, modelMapName)
+		if kdConfigMapType, ok := modelCM.Labels["kubedirectorcmtype"]; ok {
```
joel-bluedata: We shouldn't hardcode this label key string; it should be defined as a constant in one of our types.go files.

Also, the label key should be a domain-ified thing that makes use of KdDomainBase. Check some of the other label keys for examples, like ClusterAppLabel.

pkg/controller/add_kubedirectorconfigmap.go
```diff
@@ -0,0 +1,11 @@
+package controller
+
+import (
+	"github.com/bluek8s/kubedirector/pkg/controller/kubedirectorconfigmap"
```
joel-bluedata: That new controller subdirectory (and the name of this file and var names etc.) should just use "configmap", not "kubedirectorconfigmap" ... i.e. the resource we're controlling is just a "configmap" resource. Need to a search-and-replace for kubedirectorconfigmap and KubeDirectorConfigMap to fix that up.

pkg/controller/kubedirectorconfigmap/doc.go
```diff
@@ -0,0 +1,16 @@
+// Copyright 2019 Hewlett Packard Enterprise Development LP
```
joel-bluedata: 2020 :-)

pkg/validator/server.go
```diff
@@ -35,6 +35,7 @@ var validationHandlers = map[string]admitFunc{
 	"KubeDirectorApp":     admitAppCR,
 	"KubeDirectorCluster": admitClusterCR,
 	"KubeDirectorConfig":  admitKDConfigCR,
+	"ConfigMap":           admitConfigMap,
```
joel-bluedata: For this handler to be called, don't we also need a change in createAdmissionService (in validator/util.go) to register a webhook for it?

Hmm, I guess the current handler is a NOP so it doesn't matter. What were you considering for putting in that handler?

If we don't need a validator for ConfigMap then we should remove that code, and also remove it from the list of validationHandlers here.

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
joel-bluedata: Hmm I see this is also being used for reaction to kdcluster connections changes... I'll look over that in a bit.

pkg/observer/observer.go
```diff
@@ -91,6 +91,21 @@ func GetPod(
 	return result, err
 }
 
+// GetConfigMap finds the k8s Pod with the given name in the given namespace.
```
joel-bluedata: typo Pod

Gopkg.lock
```diff
@@ -0,0 +1,1193 @@
+# This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
```
joel-bluedata: Can remove this file if you're synced to master.

pkg/controller/kubedirectorconfigmap/kubedirectorconfigmap_controller.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// add adds a new Controller to mgr with r as the reconcile.Reconciler.
+func add(
+	mgr manager.Manager,
+	r reconcile.Reconciler,
+) error {
+
+	// Create a new controller
+	c, err := controller.New("kubedirectorconfigmap-controller", mgr, controller.Options{Reconciler: r})
```
joel-bluedata: We should allow multiple reconciler handlers for configmaps. See the kdcluster controller for an example -- in that case we allow 10 concurrent handlers. 10 seems like a fine number here too, which would at least help shake out whether there are any issues with concurrent handlers. Not sure we have any basis yet for picking some other number.

pkg/controller/kubedirectorconfigmap/kubedirectorconfigmap_controller.go
```diff
@@ -0,0 +1,121 @@
...
+
+// blank assignment to verify that ReconcileKubeDirectorConfigMap implements
+// reconcile.Reconciler.
+var _ reconcile.Reconciler = &ReconcileKubeDirectorConfigMap{}
+
+const (
+	// Period between the time when the controller requeues a request and
+	// it's scheduled again for reconciliation. Zero means don't poll.
+	// For now we don't need polling on this CR. Some anticipated features
+	// will need polling at which point we can change this.
```
joel-bluedata: Should remove the comment about "some anticipated features" (unless it is true!).

Do we really not need polling in this case? I guess I can see that.

joel-bluedata: Also, to be picky, this is not a CR (custom resource). Just a (standard) resource!

I have the feeling I'm also going to see "CR" in a lot of the new variable names that represent a configmap. Ideally we'd change those to just say configmap instead. Don't know how much of a hassle that will be.

pkg/controller/kubedirectorconfigmap/kubedirectorconfigmap_controller.go
```diff
@@ -0,0 +1,121 @@
...
+	// Fetch the KubeDirectorConfigMap instance.
+	cr := &corev1.ConfigMap{}
+	err := shared.Get(context.TODO(), request.NamespacedName, cr)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after
+			// reconcile request. Owned objects are automatically garbage
+			// collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue.
+			//shared.RemoveGlobalConfig()
```
joel-bluedata: This commented-out part seems like a C&P error.

pkg/controller/kubedirectorconfigmap/kubedirectorconfigmap_controller.go
```diff
@@ -0,0 +1,121 @@
...
+var _ reconcile.Reconciler = &ReconcileKubeDirectorConfigMap{}
+
+const (
+	// Period between the time when the controller requeues a request and
+	// it's scheduled again for reconciliation. Zero means don't poll.
+	// For now we don't need polling on this CR. Some anticipated features
+	// will need polling at which point we can change this.
+	reconcilePeriod = 0
+
+//reconcilePeriod = 0
```
joel-bluedata: Don't need this commented-out line.

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
joel-bluedata: OK I see what you're doing there.

For discussion's sake:

There's a slightly more expensive way to handle this change detection that avoids the ugliness of adding something to the spec (which ideally only the enduser client would modify) and also avoids having to store all the connections info in the status stanza. Something I saw mentioned in a talk at KubeCon.

We could have a property in the status that is a hash of the connections. You can use DeepHashObject from k8s.io/kubernetes/pkg/util/hash ... probably other methods. So the validator wouldn't need to write anything to the spec, but every time through the reconciler the reconciler would need to do this hash and compare it to what is stored in the status.

I'll think further & sleep on this, and also on the question of whether the clusters/configmaps should be wrapped in a Connections object together or not. (You too!)

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -230,14 +232,64 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
...
+	if state == clusterMembersChangedUnready || (cr.Spec.ConfigMetaGenerator != cr.Status.LastConfigMetaGenerator) {
 		cr.Status.SpecGenerationToProcess = &cr.Generation
 	}
 	membersErr := syncMembers(reqLogger, cr, roles, configmetaGen)
 	if membersErr != nil {
 		errLog("members", membersErr)
 		return membersErr
 	}
+	cr.Status.LastConfigMetaGenerator = cr.Spec.ConfigMetaGenerator
+	cr.Status.State = string(clusterReady)
```
joel-bluedata: We cannot set clusterReady status here. We may not be done processing changes on a single handler pass.

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -230,14 +232,64 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
...
+	// If configMetaGenerator is not incremented then return
```
joel-bluedata: If you want to do all this subsequent work at the point when a cluster transitions from not-ready to ready, you should hook into the code above where we log the "stable" event.

Also I think there's a little redundancy or confusion here stemming from tracking both cluster connections and configmaps with the same generation number. I.e. if one of those changes we still end up doing all the work as if both had changed. That may mean we should track these things with two different generation numbers (or two different hashes as per my other comment).

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
joel-bluedata: Thinking now about the case where some cluster-I'm-connected to changes; if we were doing the hash approach, what would trigger me to update my connections? Currently you have the reconciler for the changed cluster doing a write back to my spec.

In the hash approach, we'd have to include something from each cluster-we're-connected to in the hash. I think we'd only need to include the specGenerationToProcess (from the status stanza). Granted we'd have to read each connected kdcluster object to find it out, but that shouldn't be a LOT of clusters.

I know in your current code you also are trying to take care to only propagate changes from connected clusters once the connected cluster reaches a ready status, I think there would be ways to deal with that in this approach but I will hold off for now on going further down this rabbit hole.

pkg/controller/kubedirectorconfigmap/configmap.go
```diff
@@ -0,0 +1,218 @@
...
+	kdv1 "github.com/bluek8s/kubedirector/pkg/apis/kubedirector/v1beta1"
+	"github.com/bluek8s/kubedirector/pkg/observer"
+	"github.com/bluek8s/kubedirector/pkg/shared"
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var (
+	// StatusGens is exported so that the validator can have access
+	// to the KubeDirectorConfig CR StatusGens
```
joel-bluedata: comment is wrong

pkg/controller/kubedirectorconfigmap/configmap.go
```diff
@@ -0,0 +1,218 @@
...
+// change in or addition of configmap instance, currently there is no
+// polling for this resource. If the configmap is not labeled
+// with key "kubedirectorcmtype" then no op
+func (r *ReconcileKubeDirectorConfigMap) syncConfigMap(
+	reqLogger logr.Logger,
+	cr *corev1.ConfigMap,
+) error {
+	// Memoize state of the incoming object.
+	oldMap, _ := observer.GetConfigMap(cr.Namespace, cr.Name)
+	hadFinalizer := shared.HasFinalizer(cr)
```
joel-bluedata: It doesn't look like we have any need for finalizers.

Except if the configmap is removed? This code isn't currently doing anything on configmap removal (I think) but what would the desired behaviour be? Do we just notify (update) any kdclusters that have connected to the configmap?

joel-bluedata:
Done with review pass. I probably could have dug into a few more details of the reconcilers, but I want to first settle in my head the spec generation approach vs. the hash approach.

joel-bluedata:
I'll also look at the other kinds of reconciler handler triggering that we can do now that we are up to a more recent version of the operator SDK.

Cf. https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/handler

We currently use EnqueueRequestForObject.

I don't think we can make use of EnqueueRequestForOwner, since we don't necessarily want to mark our kdclusters as owners of the configmaps they consume.

EnqueueRequestsFromMapFunc is interestingly flexible... the idea is that when a change happens to object of type A, we provide a function that examines A and can trigger reconciliation requests for other objects. On the face of it this seems super useful since one of the things we were pushing against was how to trigger reconciliation in object A when object B changes. I'll need to look into this a bit more.

What I think we should do with this PR is:

* Keep the current reconciliation triggering approach, i.e. using a property in the spec that needs to be modified when we want to externally trigger a reconcile for the object, and a corresponding tracking property in the status. We may need to split these in two, to reflect the fact that we are dealing both with configmap updates and cluster updates, as mentioned in a comment above. But otherwise we can keep these the same for now.

* Fix any other issues that arise. Make sure that the underlying processing (once a reconciliation is triggered) is rock solid. I think it looks pretty close, just some comments above.

* Make sure we have the representation we want in terms of the CR schema and the in-container JSON.

* Settle any initial issues around introducing v1beta2 so that we can include these CR schema changes.

Once we nail those items, we can merge this. In the meantime I can look into whether there is a better way to trigger and handle reconciliation without having to add those generation tracking properties to the spec. E.g. the things I've mentioned about hashes and maybe making use of EnqueueRequestsFromMapFunc. Depending on what else is on your plate, you could be looking at that too.

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
kmathur2: Hmm, I thought about this. Even if we put config maps outside of the connection we still support only specific k8s resources. This is just a way to express that some resources could be "connected" to cluster and we can then act/watch. We want to be able to communicate that certain resources could be "connected" to a cluster.  With or without the wrapper the data structures need to be adjusted accordingly.

pkg/apis/kubedirector/v1beta1/kubedirectorapp_types.go
kmathur2: yes. As mentioned below. Will do this in a later PR

pkg/apis/kubedirector/v1beta1/kubedirectorapp_types.go
kmathur2: Yea got removed while resolving merge conflicts I guess. Will add back

pkg/controller/add_kubedirectorconfigmap.go
kmathur2: Yup, will do

pkg/validator/server.go
kmathur2: Yes, I was doing some stuff there. Will remove now

pkg/controller/kubedirectorconfigmap/kubedirectorconfigmap_controller.go
kmathur2: Yes, makes sense

pkg/controller/kubedirectorcluster/cluster.go
kmathur2: For the first point, we want to do this just after a new config meta is generated for this cluster. If we move this in the block above then after re-generating config meta, we do this in the next run of reconciler and if by then this cluster is updated again then we would lose that.

kmathur2: #1 Done. For the other point we need to redesign using hashes. Leaving for next PR

kmathur2:
> I'll also look at the other kinds of reconciler handler triggering that we can do now that we are up to a more recent version of the operator SDK.
> 
> Cf. https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/handler
> 
> We currently use EnqueueRequestForObject.
> 
> I don't think we can make use of EnqueueRequestForOwner, since we don't necessarily want to mark our kdclusters as owners of the configmaps they consume.
> 
> EnqueueRequestsFromMapFunc is interestingly flexible... the idea is that when a change happens to object of type A, we provide a function that examines A and can trigger reconciliation requests for other objects. On the face of it this seems super useful since one of the things we were pushing against was how to trigger reconciliation in object A when object B changes. I'll need to look into this a bit more.
> 
> What I think we should do with this PR is:
> 
> * Keep the current reconciliation triggering approach, i.e. using a property in the spec that needs to be modified when we want to externally trigger a reconcile for the object, and a corresponding tracking property in the status. We may need to split these in two, to reflect the fact that we are dealing both with configmap updates and cluster updates, as mentioned in a comment above. But otherwise we can keep these the same for now. **DONE**
> * Fix any other issues that arise. Make sure that the underlying processing (once a reconciliation is triggered) is rock solid. I think it looks pretty close, just some comments above. DONE
> * Make sure we have the representation we want in terms of the CR schema and the in-container JSON.  **Hmm Not too sure what should be changing here, based on the current MLOPs apps this has everything we need, we might have to revisit it later**
> * Settle any initial issues around introducing v1beta2 so that we can include these CR schema changes. -> We can settle this only after we settle on configmaps to be used as is or as mounted also, for the hash approach we will need to think if we should keep connectable items in a wrapper structure like connections or not so we will have to settle this in next PR.
> 
> Once we nail those items, we can merge this. In the meantime I can look into whether there is a better way to trigger and handle reconciliation without having to add those generation tracking properties to the spec. E.g. the things I've mentioned about hashes and maybe making use of EnqueueRequestsFromMapFunc. Depending on what else is on your plate, you could be looking at that too.

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -1081,12 +1086,17 @@ func queueNotify(
...
 		deltaFqdns = fqdnsList(cr, creatingOrCreated)
 	}
+
 	if op == "" {
 		if deletePending, ok := modifiedRole.membersByState[memberDeletePending]; ok {
 			op = "delnodes"
 			deltaFqdns = fqdnsList(cr, deletePending)
+		} else {
+			stateDetail.LastSetupGeneration = stateDetail.LastConfigDataGeneration
+			return
```
joel-bluedata, line 1096: I don't think we need/want to change queueNotify anymore, now that syncMemberNotifies has been changed.

deploy/example_clusters/cr-cluster-centos7.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: "centos7"
 spec:
   app: centos7x
+  connections:
```
joel-bluedata, line 7: ?

pkg/apis/kubedirector/v1beta1/kubedirectorapp_types.go
```diff
@@ -16,23 +16,24 @@ package v1beta1
 
 import (
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
joel-bluedata, line 19: This addition is wrong/redundant.

pkg/catalog/configmeta.go
```diff
@@ -105,9 +115,19 @@ func servicesForRole(
 					FQDNs: refkeys{
 						BdvlibRefKey: []string{"nodegroups", "1", "roles", roleName, "fqdns"},
 					},
-					ExportedService: "", // currently, always empty
+					ExportedService: serviceDef.ExportedService, // currently, always empty
```
joel-bluedata, line 118: comment should be deleted

pkg/catalog/types.go
```diff
@@ -42,6 +43,23 @@ type nodegroup struct {
...
+	Clusters   map[string]clusterConnections           `json:"clusters"`
+	ConfigMaps map[string]map[string]map[string]string `json:"configmaps"`
+}
+
+type clusterConnections struct {
+	Version    string                  `json:"version"`
+	Services   map[string]ngRefkeysMap `json:"services"`
+	Nodegroups map[string]nodegroup    `json:"nodegroups"`
+	Distros    map[string]refkeysMap   `json:"distros"`
+	Cluster    cluster                 `json:"cluster"`
```
joel-bluedata, line 56: The Cluster object already has Name, Isolated, ID, and ConfigMeta in it, so this seems redundant.

This seems like it should be exactly like the configmeta type above, except we would always leave Connections empty. Do we need to define these as two different types?

pkg/apis/kubedirector/v1beta1/kubedirectorapp_types.go
```diff
@@ -27,7 +27,7 @@ type KubeDirectorAppSpec struct {
 	SchemaVersion       int                 `json:"configSchemaVersion"`
 	DefaultImageRepoTag *string             `json:"defaultImageRepoTag,omitempty"`
 	DefaultSetupPackage SetupPackage        `json:"defaultConfigPackage,omitempty"`
-	Services            []Service           `json:"services,omitempty"`
+	Services            []Service           `json:"services"`
```
joel-bluedata, line 30: This change intentional?

pkg/catalog/configmeta.go
```diff
@@ -116,6 +136,83 @@ func servicesForRole(
...
+// configmaps to be connected to the given cluster
+func genconfigConnections(
+	cr *kdv1.KubeDirectorCluster,
+) (map[string]map[string]map[string]string, error) {
+
+	cmMap := make(map[string]map[string]string)
+	kdcm := make(map[string]map[string]map[string]string)
+	for _, connectedCmName := range cr.Spec.Connections.ConfigMaps {
+		cm, err := observer.GetConfigMap(cr.Namespace, connectedCmName)
+		if kdConfigMapType, ok := cm.Labels["kubedirectorcmtype"]; ok {
```
joel-bluedata, line 150: We shouldn't hardcode this label key string; it should be defined as a constant in one of our types.go files.

Also, the label key should be a domain-ified thing that makes use of KdDomainBase. Check some of the other label keys for examples, like ClusterAppLabel.

pkg/controller/configmap/configmap_controller.go
```diff
@@ -0,0 +1,116 @@
...
+	// Fetch the configmap instance.
+	cr := &corev1.ConfigMap{}
+	err := shared.Get(context.TODO(), request.NamespacedName, cr)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after
+			// reconcile request. Owned objects are automatically garbage
+			// collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue.
+			//shared.RemoveGlobalConfig()
```
joel-bluedata, line 106: This commented-out line seems like a C&P error.

pkg/validator/cluster.go
```diff
@@ -911,6 +938,13 @@ func admitClusterCR(
 		}
 	}
 
+	// If connections, be it cluster or configmap
+	// is updated then increment configMetaGenerator to trigger
+	// regeneration of configmeta
+	if !reflect.DeepEqual(clusterCR.Spec.Connections, prevClusterCR.Spec.Connections) {
+		clusterCR.Spec.ConfigMetaGenerator = prevClusterCR.Spec.ConfigMetaGenerator + 1
+	}
```
joel-bluedata, line 946: I don't think we need to change the in-memory CR here... none of the following validator code is going to look at it. The new validateConnections function you added will take care of patching the resource stored in etcd.

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -206,9 +206,57 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
...
+			for _, kubecluster := range allClusters.Items {
+				if amIBeingConnectedToThis(kubecluster) {
+					shared.LogInfof(
+						reqLogger,
+						cr,
+						shared.EventReasonConfigMap,
+						"cluster {%s} is connected to another cluster {%s} updating its configmeta",
+						cr.Name,
+						kubecluster.Name,
+					)
```
joel-bluedata, line 235: This log message is a little ambiguous/misleading about the direction of the connection. As far as our representation goes, it's kubecluster.Name that "is connected to" cr.Name.

Also, it seems like we would want to log this event on the kubecluster object, not on the cr object?

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -443,6 +491,11 @@ func (r *ReconcileKubeDirectorCluster) handleNewCluster(
 			shared.EventReasonNoEvent,
 			"ignoring cluster CR with stale status UID; will retry",
 		)
+		// If ConfigMetaGenerator has been updated then ignore UID check and
+		// return
+		if cr.Spec.ConfigMetaGenerator != cr.Status.LastConfigMetaGenerator {
+			return true, nil
+		}
```
joel-bluedata, line 498: This alarms me a little bit. Were you seeing the mismatch error message a lot?

pkg/controller/configmap/configmap.go
```diff
@@ -0,0 +1,211 @@
...
+	reqLogger logr.Logger,
+	cr *corev1.ConfigMap,
+) error {
+	// Memoize state of the incoming object.
+	oldMap, _ := observer.GetConfigMap(cr.Namespace, cr.Name)
+
+	// Set a defer func to write new status and/or finalizers if they change.
+	defer func() {
+
+		if _, ok := oldMap.Labels["kubedirectorcmtype"]; !ok {
```
joel-bluedata, line 49: same comment about label key

pkg/controller/configmap/configmap.go
```diff
@@ -0,0 +1,211 @@
...
+		for _, kubecluster := range allClusters.Items {
+			if isClusterUsingConfigMap(cr.Name, kubecluster) {
+				shared.LogInfof(
+					reqLogger,
+					cr,
+					shared.EventReasonConfigMap,
+					"configmap {%s} is connected to cluster {%s} updating its configmeta",
+					cr.Name,
+					kubecluster.Name,
+				)
```
joel-bluedata, line 74: similar comments/questions about direction of connection and where to log the event

pkg/controller/configmap/configmap.go
```diff
@@ -0,0 +1,211 @@
...
+// handleNewConfig looks in the cache for the last-known status generation
+// UID for this CR. If there is one, make sure the UID is what we expect, and
+// if so return true to keep processing the CR. If there is not any last-known
+// UID, this is either a new CR or one that was created before this KD came up.
+// In the former case, where the CR status itself has no generation UID: set
+// the config state to creating (this will also trigger population of the
+// generation UID) and return false to cause this handler to exit; we'll pick
+// up further processing in the next handler. In the latter case, sync up our
+// internal state with the visible state of the CR and return true to continue
+// processing.
```
joel-bluedata, line 130: These comments about status generation don't apply here. Also in general I don't know that a function like this serves any purpose for what we are doing with configmaps. There's a bunch of things in the function that puzzle me but my top-level reaction would be that we don't need it.

pkg/controller/configmap/configmap.go
```diff
@@ -0,0 +1,211 @@
...
+		incomingCM.ResourceVersion,
+	)
+	return true, nil
+}
+
+// handleFinalizers will, if deletion has been requested, try to do any
+// cleanup and then remove our finalizer from the in-memory CR. If deletion
+// has NOT been requested then it will add our finalizer to the in-memory CR
+// if it is absent.
+func (r *ReconcileConfigMap) handleFinalizers(
```
joel-bluedata, line 187: Do we need/use finalizers for configmaps at all?

joel-bluedata:
Done w/ review pass. I think that some of my earlier comments above are still unresolved, but I'll try to recap those here (or in some cases I just posted it again on the relevant code).

So: along with my new comments above, many of which are relatively minor, I think we have these remaining concerns:

* Several questions around how to model this, all kind of circling around 2 things: 1) whether the clusters and configmaps relationships are "the same sort of thing", and 2) how much consistency we should have with the already-implemented secrets handling.
  * Should clusters and configmaps be bundled together as "connections"?
  * Should cluster and/or configmap relationships be specifiable at per-role granularity?
  * Should configmaps be mounted rather than injected? (Probably not, since mounts can't be live-modified.)

* Some things related to the new properties in spec and status... probably should be addressed even if these are temporary properties.
  * The kdcluster CRD needs to be updated. Soon we need to tackle how to represent v1beta2 (Koshik will also be interested in this) but we could stick the changes in v1beta1 for now.
  * The "ConfigMetaGenerator" property names need changing. "ConfigMeta" terminology is too inside-baseball and also clashes with another thing in our CRs already called configmeta.

* Still some usage of "cr" or "CR" in variable names that refer to configmap objects (which are not CRs), in the pkg/controller/configmap/* files.

* Minor thing: I see several "Copyright 2019" lines being added. Should be 2020.


I'll email you about that first point on modelling, rather than working it out here on Github.

pkg/validator/cluster.go
kmathur2: If I remember this correctly then, if a new configmap is added or if an existing configmap is deleted to the connections then we would need this

pkg/controller/kubedirectorcluster/cluster.go
kmathur2: Yes there is mismatch error otherwise

joel-bluedata: When does the error happen? When you edit a kdcluster to change its connections? Or when you edit some other cluster or configmap that this one is connected to?

This particular case in this function means that a cluster CR is coming in with a gen UUID in the status that is different from the value we last wrote there. This is a worrying thing because we should be the only entity allowed to write to the status. The only way it can happen in normal operation is when the read cache sends us an old version of the object, which doesn't yet reflect the changes we most recently wrote to it. If this happens we _must_ drop the object without processing it, it can cause real shenanigans otherwise.

pkg/validator/cluster.go
joel-bluedata: The value written to here is not used by anything.

kmathur2: Yeah , its not required i guess

pkg/controller/kubedirectorcluster/cluster.go
kmathur2: I guess this was coming due to extensive re-deploying in my last round of testing, I also fixed certain sequencing since the last testing and not seeing this anymore. So I will remove this. Thanks for pointing this out.

pkg/controller/configmap/configmap.go
kmathur2: Changed language slightly and also added log in the kubecluster

pkg/controller/kubedirectorcluster/cluster.go
kmathur2: Added

joel-bluedata:
Did a scrub back through to resolve the low-hanging-fruit comments. I just need to look over the configmap and cluster reconcilers one more time now.

joel-bluedata:
Still got an open issue/question about the event logging that happens in syncConfigMap and in syncCuster.

In both of these cases, some object foo (a configmap or a cluster) has changed, and so we need to perform some operation on object bar (a cluster that foo is connected to).

The current code is posting the event on object foo. But it seems to me that since we are modifying object bar, we want to post it on object bar, as an indicator of why object bar is being changed by us (the operator).

I guess there's also an argument for post it on object foo, to indicate "hey because this changed FYI something else is going to change". I could go either way on that. But I definitely think we need to post it on object bar.

joel-bluedata:
The kdcluster CRD needs to be updated with the new spec and status properties.

pkg/controller/configmap/configmap.go
```diff
@@ -0,0 +1,108 @@
...
+	// ConfigMapType is a label placed on desired comfig maps that
+	// we want to watch and propogate inside containers
+	configMapType = shared.KdDomainBase + "/cmType"
+)
+
+var (
+	// StatusGens is exported so that the validator can have access
+	// to the ConfigMap StatusGens
+	StatusGens = shared.NewStatusGens()
+)
```
joel-bluedata, line 39: This is not used.

pkg/controller/configmap/configmap.go
```diff
@@ -0,0 +1,108 @@
...
+
+// syncConfigMap runs the reconciliation logic. It is invoked because of a
+// change in or addition of configmap instance, currently there is no
+// polling for this resource. If the configmap is not labeled
+// with key ConfigMapType then no op
+func (r *ReconcileConfigMap) syncConfigMap(
+	reqLogger logr.Logger,
+	configmap *corev1.ConfigMap,
+) error {
```
joel-bluedata, line 48: Need a blank line right after this.

pkg/controller/configmap/configmap.go
```diff
@@ -0,0 +1,108 @@
...
+	// Memoize state of the incoming object.
+	oldMap, _ := observer.GetConfigMap(configmap.Namespace, configmap.Name)
+
+	// Set a defer func to write new status and/or finalizers if they change.
```
joel-bluedata, line 52: This comment isn't right.

(Also there's no need to have this in a defer func at this point.)

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -206,9 +206,57 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
...
+						"cluster {%s} is a connection to another cluster, updating {%s}'s configmeta",
+						cr.Name,
+						kubecluster.Name,
+					)
+					updateMetaGenerator := &kubecluster
+					updateMetaGenerator.Spec.ConnectionsGenToProcess = kubecluster.Spec.ConnectionsGenToProcess + 1
+					//Notify cluster by incrementing configmetaGenerator
+					wait := time.Second
+					maxWait := 4096 * time.Second
+					for {
```
joel-bluedata, line 241: The process of
* read cluster object
* increment the observed gen number in spec
* write cluster object

Needs to be inside this for loop. That's because the most likely reason for this update to fail is because of a conflict with some other write... so, we will need to read the cluster object again to get a copy of the object that includes the new updated resource version.

Also if we want to guarantee that we actually are bumping the spec gen by 1, then that's another reason to have the whole read-modify-write process inside the loop here.

As a side note: you COULD sidestep this looping by using PATCH, now that we have moved to a new version of the SDK that supports PATCH. However if you don't want to dig into that right now I don't blame you. I think it's more straightforward at the moment just to move the read-modify-write inside the for loop.

pkg/controller/configmap/configmap.go
```diff
@@ -0,0 +1,108 @@
...
+				updateMetaGenerator := &kubecluster
+				updateMetaGenerator.Spec.ConnectionsGenToProcess = kubecluster.Spec.ConnectionsGenToProcess + 1
+				//Notify cluster by incrementing configmetaGenerator
+				wait := time.Second
+				maxWait := 4096 * time.Second
+				for {
+					if shared.Update(context.TODO(), updateMetaGenerator) == nil {
+						break
+					}
+					if wait > maxWait {
```
joel-bluedata, line 90: Similar comment about the for loop here and read-modify-write.

joel-bluedata:
Done w/ review pass. We're in the home stretch!

kmathur2:
> Still got an open issue/question about the event logging that happens in syncConfigMap and in syncCuster.
> 
> In both of these cases, some object foo (a configmap or a cluster) has changed, and so we need to perform some operation on object bar (a cluster that foo is connected to).
> 
> The current code is posting the event on object foo. But it seems to me that since we are modifying object bar, we want to post it on object bar, as an indicator of why object bar is being changed by us (the operator).
> 
> I guess there's also an argument for post it on object foo, to indicate "hey because this changed FYI something else is going to change". I could go either way on that. But I definitely think we need to post it on object bar.

We are doing it on cluster (bar)	

if cr.Status.LastConnectionGen != cr.Spec.ConnectionsGenToProcess {
			shared.LogInfof(
				reqLogger,
				cr,
				shared.EventReasonCluster,
				"regenerating in-cluster configmeta due to chamge in connections",
			)
			cr.Status.LastConnectionGen = cr.Spec.ConnectionsGenToProcess
		}

joel-bluedata:
Ah I see. Sorry! Still, it might be good to have it say which connection changed, which might be easier to do in the other spot.

BTW typo "chamge" there.

kmathur2:
> Ah I see. Sorry! Still, it might be good to have it say which connection changed, which might be easier to do in the other spot.
> 
> BTW typo "chamge" there.

We are doing in both places, example - 

{"level":"info","ts":1587149669.1363883,"logger":"controller_configmap","msg":"configmap {model1} is a connection to cluster {flask-and-inf-eng}, updating its configmeta","Request.Namespace":"c1","Request.Name":"model1"}
{"level":"info","ts":1587149669.1621995,"logger":"controller_kubedirectorcluster","msg":"regenerating in-cluster configmeta due to chamge in connections","Request.Namespace":"c1","Request.Name":"flask-and-inf-eng"}

kmathur2:
> The kdcluster CRD needs to be updated with the new spec and status properties.

Won't this go in v1Beta2 CRD when we finalize that?

joel-bluedata:
Eh, yeah. IMO it's nice to express it here too though, just so this whole change is self-consistent. Not a big deal if you'd rather wait (assuming we get the v1beta2 change in soon).

joel-bluedata:
> We are doing in both places

We're posting an event against foo that says foo is affecting bar, and we're posting an event against bar that says "I'm being affected by... something". What I'm suggesting is that the event on bar should ideally say that foo caused the change.

kmathur2:
> > We are doing in both places
> 
> We're posting an event against foo that says foo is affecting bar, and we're posting an event against bar that says "I'm being affected by... something". What I'm suggesting is that the event on bar should ideally say that foo caused the change.

But bar doesn't know the reason, it will re-generate the config meta because the counter was incremented. The bar reconciler does not know what exactly was changed.

joel-bluedata:
That's why I suggested posting that event in the other code.

kmathur2:
> That's why I suggested posting that event in the other code.

At the other place (foo) already have -    "configmap {%s} is a connection to cluster {%s}, updating its configmeta" , does this not suffice !!! I am soo confused

pkg/controller/configmap/configmap.go
```diff
@@ -0,0 +1,107 @@
...
+				configmap.Name,
+				kubecluster.Name,
+			)
+
+			//Notify cluster by incrementing configmetaGenerator
+			wait := time.Second
+			maxWait := 4096 * time.Second
+			for {
+				// Get a fresh copy of this cluster to work with and
+				// try update
```
joel-bluedata, line 79: The first time through this loop (which will probably succeed) we do already have a copy we just loaded. We only need to re-load and re-apply our spec inc if the Update fails. That will cut the number of GETs by 1/2 in the common case.

Same comment for the other similar loop.

joel-bluedata: I guess to be clear, what I'm saying is:

- coming into this code here we already have kubecluster
- increment its spec gen number
- try the shared.Update
- if that succeeds, exit the loop
- otherwise do the shared.Get to load a fresh copy into kubecluster and loop back around

joel-bluedata: Another thing to be aware of is that the Get could fail. If it fails because the cluster is gone, no biggy, just exit the loop. If it fails for other reasons I guess we should keep looping until maxWait is exceeded.

FWIW the status-update loop in the defer func in syncCluster has to deal with basically this same sort of flow.

kmathur2: Yes, already took care of above 2, there is a race between changes and reviewing :)

joel-bluedata: :+1:

I see that in the GET error case (other than not-found) you are returning the error rather than looping to try again. Thinking about that at the moment. I guess the question is how persistent we want to be vs. willing to drop an update. (Not that a GET error is very likely.)

joel-bluedata: 👍 👍 👍

joel-bluedata:
Taking event-posting-chat to Slack for a bit. :-)

pkg/controller/configmap/configmap.go
```diff
@@ -66,8 +66,14 @@ func (r *ReconcileConfigMap) syncConfigMap(
 				reqLogger,
 				configmap,
```
joel-bluedata, line 67: This arg should be kubecluster.

kmathur2: why will syncConfigMap take kubecluster??

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -228,8 +228,8 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 					shared.LogInfof(
 						reqLogger,
 						cr,
-						shared.EventReasonConfigMap,
-						"cluster {%s} is a connection to another cluster, updating {%s}'s configmeta",
+						shared.EventReasonCluster,
+						"connected cluster {%s} has changed; connected to cluster {%s}; updating it",
 						cr.Name,
 						kubecluster.Name,
 					)
```
joel-bluedata, line 235:
```erlang
                    shared.LogInfof(
                        reqLogger,
                        cr,
                        shared.EventReasonCluster,
                        "connected to cluster {%s}; updating it",
                        kubecluster.Name,
                    )
                    shared.LogInfof(
                        reqLogger,
                        kubecluster,
                        shared.EventReasonCluster,
                        "connected cluster {%s} has changed",
                        cr.Name,
                    )
```

joel-bluedata:
Once those pesky events are sorted I think we're good to go. I'll do a test build before merging to master... let me know if you have some example cluster CRs I can play with.

pkg/controller/configmap/configmap.go
joel-bluedata: The second argument to LogInfof is the object to post the event to.

joel-bluedata:
The comment above about event posting in configmap.go is still unresolved FYI.

Issue #273, opened 2020-02-15
In app CR, specify which roles need which lifecycle events


joel-bluedata:
A good and important improvement would be to have the KD app spec explicitly state which roles need which notifies. So if for example the workers don't need to receive add/delete notifies, we know that we don't need to update their config or send them notifies on add/delete events... and so we won't get blocked from processing KD cluster changes if some of those workers are down.

PR #299, opened 2020-04-15
BDP 930 changes

koshiksamota10:
Github issue: https://github.com/bluek8s/kubedirector/issues/273

### DIFF
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
@@ -126,6 +126,11 @@ spec:
                     items:
                       type: string
                       pattern: '^/.*[^/]$'
+                  eventList:
+                    type: array
+                    items:
+                      type: string
+                      pattern: '^/.*[^/]$'
                   minResources:
                     type: object
                     nullable: true
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
@@ -130,7 +130,7 @@ spec:
                     type: array
                     items:
                       type: string
-                      pattern: '^/.*[^/]$'
+                      pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
                   minResources:
                     type: object
                     nullable: true
@@ -185,6 +185,11 @@ spec:
               items:
                 type: string
                 pattern: '^/.*[^/]$'
+            defaultEventList:
+              type: array
+              items:
+                type: string
+                pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
             capabilities:
               type: array
               items:
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
@@ -130,7 +130,7 @@ spec:
                     type: array
                     items:
                       type: string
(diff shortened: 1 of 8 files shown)


deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
```diff
@@ -126,6 +126,11 @@ spec:
                     items:
                       type: string
                       pattern: '^/.*[^/]$'
+                  eventList:
+                    type: array
+                    items:
+                      type: string
+                      pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
```
joel-bluedata, line 133: Instead of specifying a pattern here we should enumerate the possible values. As an example of doing that, cf. the serviceType property in the kdcluster CRD.

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
```diff
@@ -180,6 +185,11 @@ spec:
               items:
                 type: string
                 pattern: '^/.*[^/]$'
+            defaultEventList:
+              type: array
+              items:
+                type: string
+                pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
```
joel-bluedata, line 192: similar comment about enumeration

joel-bluedata: Also of course, once this PR is merged we then need to look at how to express this as v1beta2 rather than just modifying the already-released v1beta1 ... we have been talking about this offline but I wanted to record it here too.

https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/

joel-bluedata:
Looking good! Let me know a bit more about your testing process.

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
koshiksamota10: I changed this.

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
koshiksamota10: Changed this. Kartik and I will come up with a plan for the v1beta2 plan.

koshiksamota10:
I have been testing by changing the spark app config file by adding various combinations of configure, addnodes and delnodes to the worker and jupyter roles. I have been keeping track of the logs by using kubectl logs. In appConfig part of the code, I look for the trace "Running app config" statement and in queueNotify code path, I look for "will notify member" logs. For the cases where we explicitly mention in the app catalog files, it would print and not otherwise. I have been adding and deleting nodes from the cluster by modifying the cluster config files.

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -1043,6 +1043,21 @@ func appConfig(
...
+			reqLogger,
+			appErr,
+			cr,
+			shared.EventReasonCluster,
+			"app referenced by cluster does not exist",
+		)
+	}
+	role := catalog.GetRoleFromID(appCr, roleName)
+	if role.EventList != nil && !shared.StringInList("configure", *role.EventList) {
+		return false, appErr
```
joel-bluedata, line 1059: I'm curious about returning false here (for the "result is final" boolean). Previously you were returning true, which seemed correct. What caused the change?

Also you could/should return nil instead of appErr (for that second value) at this point.

joel-bluedata:
Looking good I think. I will test it out before merging to master. I'd also like to take a look at these current apps to see which other ones don't need use all the events in their startscripts.

joel-bluedata:
Looks good. I'll file a separate issue for updating app CRs to take advantage of this.

Issue #263, opened 2020-02-05
GPU visible in non-GPU-requesting pod


joel-bluedata:
See https://github.com/NVIDIA/k8s-device-plugin/issues/87

This is not "our issue" per se but we can help work around it.

Simplest thing to do might be to include that "NVIDIA_VISIBLE_DEVICES=VOID" environment variable setting in our generated PodTemplateSpecs for KDCluster roles that do not request GPU resources.

attn @swamibluedata and @saurabh-jogalekar

joel-bluedata:
Also we should double-check what K8s is doing on multi-GPU hosts, when scheduling a pod that requests some smaller number of GPUs. Is it setting things up correctly so that the pod only sees the number of GPUs it asked for?

joel-bluedata:
Pushing this.

PR #302, opened 2020-04-22
Kishore test

rkishore-hpe:
This change is intended as at least a partial fix for the issue, https://github.com/bluek8s/kubedirector/issues/263.

The NVIDIA-suggested workaround (https://github.com/NVIDIA/k8s-device-plugin/issues/87), referred to by the issue's description, is implemented:
For each container pertaining to a Role that has NOT requested a GPU resource, an "NVIDIA_VISIBLE_DEVICES=VOID" environment variable is added.

We may have to watch for future updates to NVIDIA's Docker plug-in, to see if the need for this workaround is removed.

### DIFF
--- a/pkg/catalog/interrogate.go
+++ b/pkg/catalog/interrogate.go
@@ -113,7 +113,6 @@ func GetRoleCardinality(
 
 // GetRoleMinResources is a utility function that fetching the minimum resources
 // for a given app role
-// Test comment XXX
 func GetRoleMinResources(
 	appRole *kdv1.NodeRole,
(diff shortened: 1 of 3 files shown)


joel-bluedata:
Yep, looks plausible!

I remember earlier you had seen some warning messages about the operator trying to modify a non-modifiable statefulset property. I'm curious about those... wondering if this PR stimulates that in some way or if that was a coincidence that you were seeing them. Do you see those messages consistently in some situations (like when adding this env var), or is it rare/random?

PR #306, opened 2020-05-06
Cluster connections refactoring and addition of secret as connectable resource

kmathur2:
- Currently, cluster connections utilizes **connectionsGenerationToProcess** in spec to determine if connections have been updated or not. 
- It's not ideal to use a field in spec for implementation purpose and hence, removed from the spec.
- The decision of whether connection object or the actually connected resources were modified is now determined by the md5 checksum of resource versions of connected k8s resources. This checksum is stored in the cluster's status object.  
- Also added secret as one of the connectable resources similar to clusters and configmaps.
- Modified reconciler for configmap to not change the cluster's spec but to invoke the cluster's reconciler by updating annotation.
- Secret's reconciler is similar to configmap.

### DIFF
--- a/pkg/controller/kubedirectorcluster/cluster.go
+++ b/pkg/controller/kubedirectorcluster/cluster.go
@@ -311,7 +311,6 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 		return configMetaErr
 	}
 	newHash := calcConnectionsHash(&cr.Spec.Connections, cr.Namespace)
-	//fmt.Println("DEBUG: new hash: ", newHash)
 	if state == clusterMembersChangedUnready || (newHash != cr.Status.LastConnectionHash) {
 		if cr.Status.SpecGenerationToProcess == nil {
 			cr.Status.SpecGenerationToProcess = &cr.Generation
--- a/pkg/controller/kubedirectorcluster/cluster.go
+++ b/pkg/controller/kubedirectorcluster/cluster.go
@@ -334,10 +334,10 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 	return nil
 }
 
+// Calculates md5sum of resource-versions of all resources
+// connected to this cluster
 func calcConnectionsHash(con *kdv1.Connections, ns string) string {
-	//sliceToSha(buff bytes.Buffer)
 	clusterNames := con.Clusters
-	//clusterRvs := make([]string, len(clusterNames))
 	var buffer bytes.Buffer
 	for _, c := range clusterNames {
 		clusterObj, _ := observer.GetCluster(ns, c)
@@ -356,6 +356,7 @@ func calcConnectionsHash(con *kdv1.Connections, ns string) string {
 		rv := secretObj.ResourceVersion
 		buffer.WriteString(rv)
 	}
+	// md5 is very cheap for small strings
 	md5Sum := md5.Sum([]byte(buffer.String()))
 	return hex.EncodeToString(md5Sum[:])
 }
--- a/pkg/controller/kubedirectorcluster/cluster.go
+++ b/pkg/controller/kubedirectorcluster/cluster.go
@@ -288,8 +288,8 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 				}
 			}
 			cr.Status.State = string(clusterReady)
-			return nil
 		}
+		return nil
 	}
 
 	if cr.Status.State != string(clusterCreating) {
--- a/pkg/controller/kubedirectorcluster/cluster.go
+++ b/pkg/controller/kubedirectorcluster/cluster.go
@@ -17,9 +17,9 @@ package kubedirectorcluster
 import (
 	"bytes"
 	"context"
-	"crypto/sha256"
 	"fmt"
 	"reflect"
+	"strconv"
 	"time"
 
 	"sync/atomic"
@@ -246,7 +246,12 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 					updateMetaGenerator := &kubecluster
 					//updateMetaGenerator.Spec.ConnectionsGenToProcess = kubecluster.Spec.ConnectionsGenToProcess + 1
 					annotations := updateMetaGenerator.Annotations
-					annotations["kdreconfig"] = "true"
+					if v, ok := annotations["connUpdateCounter"]; ok {
+						newV, _ := strconv.Atoi(v)
+						annotations["connUpdateCounter"] = strconv.Itoa(newV + 1)
+					} else {
+						annotations["connUpdateCounter"] = "1"
+					}
 					updateMetaGenerator.Annotations = annotations
 					//Notify cluster by incrementing configmetaGenerator
 					wait := time.Second
@@ -254,7 +259,12 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 					for {
 						updateMetaGenerator := &kubecluster
 						annotations := updateMetaGenerator.Annotations
-						annotations["kdreconfig"] = "true"
+						if v, ok := annotations["connUpdateCounter"]; ok {
+							newV, _ := strconv.Atoi(v)
+							annotations["connUpdateCounter"] = strconv.Itoa(newV + 1)
+						} else {
+							annotations["connUpdateCounter"] = "1"
+						}
 						updateMetaGenerator.Annotations = annotations
 						if shared.Update(context.TODO(), updateMetaGenerator) == nil {
 							break
@@ -278,11 +288,6 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 				}
 			}
 			cr.Status.State = string(clusterReady)
-		}
-		if _, ok := cr.Annotations["kdreconfig"]; !ok {
-			return nil
-		}
-		if cr.Annotations["kdreconfig"] == "false" {
 			return nil
 		}
 	}
@@ -306,13 +311,15 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 		return configMetaErr
 	}
 	newHash := calcConnectionsHash(&cr.Spec.Connections, cr.Namespace)
+	//fmt.Println("DEBUG: new hash: ", newHash)
 	if state == clusterMembersChangedUnready || (newHash != cr.Status.LastConnectionHash) {
-		cr.Status.SpecGenerationToProcess = &cr.Generation
+		if cr.Status.SpecGenerationToProcess == nil {
+			cr.Status.SpecGenerationToProcess = &cr.Generation
+		}
 		if newHash != cr.Status.LastConnectionHash {
 			incremented := atomic.AddInt64(cr.Status.SpecGenerationToProcess, 1)
 			cr.Status.SpecGenerationToProcess = &incremented
 			cr.Status.LastConnectionHash = newHash
-			cr.Annotations["kdreconfig"] = "false"
 		}
 	}
 	membersErr := syncMembers(reqLogger, cr, roles, configmetaGen)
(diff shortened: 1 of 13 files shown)


pkg/controller/secret/secret.go
```diff
@@ -0,0 +1,123 @@
...
+
+const (
+	// secretType is a label placed on desired comfig maps that
+	// we want to watch and propogate inside containers
+	secretType             = shared.KdDomainBase + "/secretType"
+	connectionsIncrementor = shared.KdDomainBase + "/connUpdateCounter"
+)
+
+// syncSecret runs the reconciliation logic. It is invoked because of a
+// change in or addition of configmap instance, currently there is no
```
riteshja, line 39: nit: change in or addition to a secret instance?

pkg/controller/kubedirectorcluster/types.go
```diff
@@ -101,6 +102,10 @@ const (
 	zeroPortsService = "n/a"
 )
 
+const (
+	connectionsIncrementor = shared.KdDomainBase + "/connUpdateCounter"
```
joel-bluedata, line 106: Looks like this is also defined in configmap.go?

pkg/controller/secret/secret_controller.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+// ReconcileSecret reconciles a Secret object.
+type ReconcileSecret struct {
+	scheme *runtime.Scheme
+}
+
+// Reconcile reads that state of the cluster for a ConfigMap object
+// and makes changes based on the state read and what is in the
+// Secret.Spec.
```
joel-bluedata, line 87: This block of comments seems like it needs some fixin'.

joel-bluedata:
The gist of the changes looks good. I'll need to take a closer look at configmap.go, cluster.go, and secret.go (hopefully later today).

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -240,13 +246,28 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
...
+					//updateMetaGenerator.Spec.ConnectionsGenToProcess = kubecluster.Spec.ConnectionsGenToProcess + 1
+					annotations := updateMetaGenerator.Annotations
+					if v, ok := annotations[connectionsIncrementor]; ok {
+						newV, _ := strconv.Atoi(v)
+						annotations[connectionsIncrementor] = strconv.Itoa(newV + 1)
+					} else {
+						annotations[connectionsIncrementor] = "1"
+					}
+					updateMetaGenerator.Annotations = annotations
 					//Notify cluster by incrementing configmetaGenerator
```
joel-bluedata, line 258: Some comments like this need to be tweaked.

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -240,13 +246,28 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
...
-					updateMetaGenerator.Spec.ConnectionsGenToProcess = kubecluster.Spec.ConnectionsGenToProcess + 1
+					//updateMetaGenerator.Spec.ConnectionsGenToProcess = kubecluster.Spec.ConnectionsGenToProcess + 1
+					annotations := updateMetaGenerator.Annotations
+					if v, ok := annotations[connectionsIncrementor]; ok {
+						newV, _ := strconv.Atoi(v)
+						annotations[connectionsIncrementor] = strconv.Itoa(newV + 1)
+					} else {
+						annotations[connectionsIncrementor] = "1"
+					}
+					updateMetaGenerator.Annotations = annotations
```
joel-bluedata, line 257: Seems unnecessary to build the annotations here if we're just going to do it again in the top of the loop below...

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -293,10 +314,16 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 		)
 		return configMetaErr
 	}
-
-	if state == clusterMembersChangedUnready || (cr.Spec.ConnectionsGenToProcess != cr.Status.LastConnectionGen) {
-		cr.Status.SpecGenerationToProcess = &cr.Generation
-		cr.Status.LastConnectionGen = cr.Spec.ConnectionsGenToProcess
+	newHash := calcConnectionsHash(&cr.Spec.Connections, cr.Namespace)
```
joel-bluedata, line 317: If the cluster is in stable state we already calculated the hash above. I think we should try to avoid redundantly calculating the hash if possible.

We could for example cache the hash in the cluster CR, much like we cache the app CR in the cluster CR. Cf. in the definition of the KubeDirectorCluster struct type we have the field AppSpec that is flagged to keep it out of the JSON, so it's for internal use only. Whenever we want to get the app for this cluster we call catalog.GetApp which returns the cached app if available and if not finds it and caches it. Could do something very similar with this hash.

(...actually, looking at GetApp now I think that a change a while back broke the caching behavior there. Hmm, and d'oh.)

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -293,10 +314,16 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
...
+	if state == clusterMembersChangedUnready || (newHash != cr.Status.LastConnectionHash) {
+		if cr.Status.SpecGenerationToProcess == nil {
+			cr.Status.SpecGenerationToProcess = &cr.Generation
```
joel-bluedata, line 320: Curious about this part; what's the reason to initialize it to cr.Generation now?

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -306,6 +333,33 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
...
 }
 
+// Calculates md5sum of resource-versions of all resources
+// connected to this cluster
+func calcConnectionsHash(con *kdv1.Connections, ns string) string {
+	clusterNames := con.Clusters
+	var buffer bytes.Buffer
+	for _, c := range clusterNames {
+		clusterObj, _ := observer.GetCluster(ns, c)
+		rv := clusterObj.ResourceVersion
```
joel-bluedata, line 343: For clusters we might want to use specGenerationToProcess here rather than resourceVersion, as it's a more reliable indicator of whether there was a change that we care about.

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -306,6 +333,33 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
...
+		buffer.WriteString(rv)
+	}
+	secretNames := con.Secrets
+	for _, c := range secretNames {
+		secretObj, _ := observer.GetSecret(ns, c)
+		rv := secretObj.ResourceVersion
+		buffer.WriteString(rv)
+	}
+	// md5 is very cheap for small strings
+	md5Sum := md5.Sum([]byte(buffer.String()))
```
joel-bluedata, line 359: It seems like we do want to include the resource names in this hash too. We don't want to miss an update if configmap foo is switched out for configmap bar and they both happen to have the same resourceVersion.

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -306,6 +333,33 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 	return nil
 }
 
+// Calculates md5sum of resource-versions of all resources
+// connected to this cluster
+func calcConnectionsHash(con *kdv1.Connections, ns string) string {
```
joel-bluedata, line 338: formatting

```go
func calcConnectionsHash(
	con *kdv1.Connections,
	ns string,
) string {

	clusterNames := con.Clusters
```

pkg/controller/kubedirectorcluster/cluster.go
joel-bluedata: Or if these are the only two places we use the hash we don't need to get that fancy, should just put it in a local variable.

joel-bluedata:
Done w/ review pass.

pkg/controller/kubedirectorcluster/cluster.go
kmathur2: Hmm, does it matter? we can start by this being equal to Generation and go on from there.

pkg/controller/kubedirectorcluster/cluster.go
kmathur2: yes good point. Done

pkg/controller/kubedirectorcluster/cluster.go
joel-bluedata: It doesn't matter a whole lot. It does make one curious why choose to start at that number rather than at e.g. 0 or 1.

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -335,28 +326,30 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 
 // Calculates md5sum of resource-versions of all resources
 // connected to this cluster
-func calcConnectionsHash(con *kdv1.Connections, ns string) string {
-	if con == nil {
-		return ""
-	}
+func calcConnectionsHash(
+	con *kdv1.Connections,
+	ns string) string {
```
joel-bluedata, line 331:
```go
	ns string,
) string {
```

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -154,6 +154,7 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 
 	// We use a finalizer to maintain KubeDirector state consistency;
 	// e.g. app references and ClusterStatusGens.
+	var currentHash string = calcConnectionsHash(&cr.Spec.Connections, cr.Namespace)
```
joel-bluedata, line 157: Slightly confusing to have the comment about finalizers attached to this line here.

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -335,28 +326,30 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
...
 
 	clusterNames := con.Clusters
 	var buffer bytes.Buffer
 	for _, c := range clusterNames {
 		clusterObj, _ := observer.GetCluster(ns, c)
-		rv := clusterObj.ResourceVersion
-		buffer.WriteString(rv)
+		buffer.WriteString(c)
+		buffer.WriteString(strconv.Itoa(
+			int(*clusterObj.Status.SpecGenerationToProcess)))
```
joel-bluedata, line 339: I guess technically we should worry about the possibility that this is nil; don't want to get a nil deref exception here.

If it's nil we could just use the string "nil" to represent it, should be fine.

joel-bluedata:
Looking good, just a few minor open issues.

There's a separate question to discuss about how to mark the configmaps and secrets that are to be used by KD. The secret-mounting feature uses a resource name prefix, whereas here we are using labels. There's pros and cons to those methods that we should chew on at least briefly, and also some benefit to using a consistent approach. But I don't think that should be handled in this PR. I'll raise the topic with you in email or something.

kmathur2:
> Looking good, just a few minor open issues.
> 
> There's a separate question to discuss about how to mark the configmaps and secrets that are to be used by KD. The secret-mounting feature uses a resource name prefix, whereas here we are using labels. There's pros and cons to those methods that we should chew on at least briefly, and also some benefit to using a consistent approach. But I don't think that should be handled in this PR. I'll raise the topic with you in email or something.

Yes sounds good.

joel-bluedata:
FYI I've pushed an image built at this commit, bluek8s/kubedirector:03e7570

PR #307, opened 2020-05-08
Service Endpoint Auth Token generation

kmathur2:
- Added a flag in ServiceEndpoint object to determine if the endpoint needs auth token or not.
- If the above flag is set then generate auth token and store in member status as well as annotation for the k8s service object. 
- Added auth token in configmeta to be consumed by apps.

### DIFF
--- a/pkg/apis/kubedirector/v1beta1/kubedirectorapp_types.go
+++ b/pkg/apis/kubedirector/v1beta1/kubedirectorapp_types.go
@@ -95,7 +95,7 @@ type ServiceEndpoint struct {
 	Port        *int32 `json:"port"`
 	Path        string `json:"path,omitempty"`
 	IsDashboard bool   `json:"isDashboard,omitempty"`
-	AuthToken   bool   `json:"authToken,omitempty"`
+	HasToken    bool   `json:"hasToken,omitempty"`
 }
 
 // NodeRole describes a subset of virtual cluster members that will provide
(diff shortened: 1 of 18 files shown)


pkg/apis/kubedirector/v1beta1/kubedirectorapp_types.go
```diff
@@ -95,6 +95,7 @@ type ServiceEndpoint struct {
 	Port        *int32 `json:"port"`
 	Path        string `json:"path,omitempty"`
 	IsDashboard bool   `json:"isDashboard,omitempty"`
+	HasToken    bool   `json:"hasToken,omitempty"`
```
joel-bluedata, line 98: This should be authToken, not hasToken?

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
```diff
@@ -155,6 +155,7 @@ type RoleStatus struct {
 type MemberStatus struct {
 	Pod         string            `json:"pod"`
 	Service     string            `json:"service"`
+	AuthToken   string            `json:"authToken,omitempty"`
```
joel-bluedata, line 158: The kdcluster CRD needs to be updated accordingly.

pkg/apis/kubedirector/v1beta1/kubedirectorapp_types.go
joel-bluedata: Either that or the app CRD should be changed to say hasToken.

joel-bluedata:
image build at bluek8s/kubedirector:da322dc FYI

kmathur2:
> image build at bluek8s/kubedirector:da322dc FYI

👍 👍

PR #310, opened 2020-05-21
BDP 1061: CRD versioning changes.

koshiksamota10:
Changes to support newer version of CRD where some of the fields have been edited/added/deleted. New conversion webhook is invoked when objects from different versions need to conform with older/newer versions.


deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
```diff
@@ -4,198 +4,386 @@ metadata:
...
...RWlNQ0FHQTFVRUF4TVphM1ZpWldScGNtVmpkRzl5TFhaaGJHbGtZWFJ2Y2kxallUQ0NBU0l3CkRRWUpLb1pJaHZjTkFRRUJCUUFEZ2dFUEFEQ0NBUW9DZ2dFQkFOYStlc3lRVFI4S3czNXRHWjczVzNBb3FWNC8KNm9EaDh6d0lkc2h0WWt5SGFpUW1KNXgrNm9xdmRQOTVmb0lZWVp4MUcxQi9aYUpZUmlURi9udkRoOUYvZUFnTQo1VVVRSlhRUDVOM3g5TElZMWNUcVBsZDErZ2VWYjY1YjEwUjNPSkYxVDJBd0prQjlNSnNVeWJiUkQ2YUVIMGl3CnBDWFpIbmNWSVdVS29SelpzbTdmbFQxMXg3bVFqYmZGUXBrL2Z1cG5OTGw4UVNqallocUpDa2cxM1M0eFVBNGoKTFVvNGUzeEx0MGRvbTFuSUQ5czg4NkVvbzNML203bnVBMXp2MjZOMUZLUlNvYnV0SmVPbDZSM0JZMFl6a2tPNAowNHBEWkZIM2pybENrUnB4anN1SVJZbC9jY3hFZURQQkZNcUovYVlJT01QUkplVE9TdlJHQzNmTzloOENBd0VBCkFhTWpNQ0V3RGdZRFZSMFBBUUgvQkFRREFnS2tNQThHQTFVZEV3RUIvd1FGTUFNQkFmOHdEUVlKS29aSWh2Y04KQVFFTEJRQURnZ0VCQURaTzdkczEwd2tBTWtHcXhuc01FcFlkUjhmTHg2RGtrYTd2dHFoalh1QmpOOUhZZDIraAp5UDZ5eGVqTmYyWHp0STIyQzFKS21tL2JUTC9oMCtKWWcza0RLYlNhRDZveUtSOXMwY2twVmJmVkpDbGc5a3N4CjFvVmVrdk5iSFNBcVh2M0hZN3RkT01QOVpZOWtrNU1IVnVYQzRYdUpVVWFyam1PR3pOTDhvR291Yk1KZmNtclAKMjNWMmErcmIxTjRBWWdCUjY1RjNBUVUvRUJ0UDM0eDcyei9GZm5xdnppWTFOMno5RUloWG8wbDBDTjhsNG5Wbgo3cDJUSkt6RWQvOW50YWROekFSSThsUGhYV3I0UnRLWmN3Tmt0ajRNWHhHd0toMU8vZE1xWVE2a2g1bzgzQjgrCmpNNnIrbTNueFVMWmR6eTBDM1JRenV3a1NwbzROLzVhNmcwPQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==
```
kmathur2, line 381: we shouldn't have this hardcoded value here

pkg/shared/client.go
```diff
@@ -232,3 +232,11 @@ func Delete(
 
 	return client.Delete(ctx, obj, opts...)
 }
+
+func Patch(
+	ctx context.Context,
+	obj runtime.Object,
+	patch k8sClient.Patch,
+) error {
+	return client.Patch(ctx, obj, patch)
```
kmathur2, line 241: Need space here and comment for this fun

pkg/validator/util.go
```diff
@@ -86,7 +86,7 @@ func createWebhookService(
 
 // createAdmissionService creates our MutatingWebhookConfiguration resource
```
kmathur2, line 87: Need to change comment based on func name

joel-bluedata:
We probably shouldn't, or at least we don't need to, specify the defautEventList for kdapp CRs that use all the events anyway.

Also for the ones that don't have any config package at all, like the centos kdapps, it looks a bit odd to specify the event list.

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
```diff
@@ -4,198 +4,386 @@ metadata:
...
-                  type: string
-                  pattern: '^(file|https?)://.+\.tgz$'
-            services:
-              type: array
-              items:
+  preserveUnknownFields: false
+  versions:
+  - name: v1beta1
+    served: true
+    storage: true
```
joel-bluedata, line 11: Shouldn't beta2 be the storage version?

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
```diff
@@ -4,198 +4,386 @@ metadata:
...
+                          maximum: 65535
+                        urlScheme:
+                          type: string
+                          minLength: 1
+                          maxLength: 15
+                          pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
+                        path:
+                          type: string
+                        isDashboard:
+                          type: boolean
```
joel-bluedata, line 94: missing hasAuthToken here

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
```diff
@@ -4,198 +4,386 @@ metadata:
...
```
joel-bluedata, line 271: missing hasAuthToken here

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
```diff
@@ -4,198 +4,386 @@ metadata:
...
+              capabilities:
+                type: array
+                items:
+                  type: string
+                  minLength: 1
+              systemdRequired:
+                type: boolean
+  conversion:
+    strategy: Webhook
+    conversionReviewVersions: ["v1beta1", "v1beta2"]
```
joel-bluedata, line 375: This should only have v1beta1 here. This is talking about the version of the ConversionReview object I think, not the version of the CR.

pkg/validator/convertor.go
```diff
@@ -0,0 +1,216 @@
...
+
+	serializer := getInputSerializer(contentType)
+	if serializer == nil {
+		msg := fmt.Sprintf("invalid Content-Type header `%s`", contentType)
+		http.Error(w, msg, http.StatusBadRequest)
+		return
+	}
+
+	convertReview := v1beta1.ConversionReview{}
+	_, _, err := serializer.Decode(body, nil, &convertReview)
```
joel-bluedata, line 193: Why this approach as opposed to using json.Unmarshal as in the validator?

pkg/validator/convertor.go
```diff
@@ -0,0 +1,216 @@
...
+	convertReview.Request = &v1beta1.ConversionRequest{}
+
+	accept := r.Header.Get("Accept")
+	outSerializer := getOutputSerializer(accept)
+	if outSerializer == nil {
+		msg := fmt.Sprintf("invalid accept header `%s`", accept)
+		http.Error(w, msg, http.StatusBadRequest)
+		return
+	}
+	err = outSerializer.Encode(&convertReview, w)
```
joel-bluedata, line 211: similar question w.r.t. json.Marshal

pkg/validator/convertor.go
```diff
@@ -0,0 +1,216 @@
...
+				clause.Type == "*" && clause.SubType == "*":
+				return v
+			}
+		}
+	}
+
+	return nil
+}
+
+func convert(
```
joel-bluedata, line 69: Having functions named "convert" and "convertor" is a bit ambiguous, maybe tweak that a bit to be more specifically descriptive.

We also typically have function header comments for important functions even if they're not exported.

pkg/validator/convertor.go
```diff
@@ -0,0 +1,216 @@
...
+			}
+		}
+
+		switch fromVersion {
+		case "kubedirector.hpe.com/v1beta1":
+			switch toVersion {
+			case "kubedirector.hpe.com/v1beta2":
+				spec := convertedObject.Object["spec"]
+				delete(convertedObject.Object, "spec")
+				spec.(map[string]interface{})["defaultEventList"] = []string{"configure", "addnodes", "delnodes"}
```
joel-bluedata, line 104: I think when going from v1beta1 to v1beta2 we don't have to do anything to the spec, since new fields added in v1beta2 are optional and when omitted have the same behavior as v1beta1.

pkg/validator/convertor.go
```diff
@@ -0,0 +1,216 @@
...
+				delete(spec.(map[string]interface{}), "defaultEventList")
+
+				roles := spec.(map[string]interface{})["roles"]
+				delete(spec.(map[string]interface{}), "roles")
+				for _, roleConfig := range roles.([]interface{}) {
+					delete(roleConfig.(map[string]interface{}), "eventList")
+				}
+
+				spec.(map[string]interface{})["roles"] = roles
+				convertedObject.Object["spec"] = spec
```
joel-bluedata, line 128: We'll also need to handle hasAuthtToken.

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
joel-bluedata: Hmm maybe not. However we would have to be careful not to lose the new v1beta2 fields when the object is stored. I'm reading over https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md at the moment and we may want to discuss some aspects of that later.

joel-bluedata:
So if we do leave v1beta1 as the storage version, our reconciliation handlers can continue to work with v1beta1 objects... that's good. I think we'd need an addition to the Rules for the admission webhook though so that it will intercept both incoming v1beta1 and v1beta2 objects.

I'm curious how exactly pruning behaves (preserveUnknownFields false) with v1beta1 as the storage version. If you create a v1beta2 kdapp with for example defaultEventList set to something, and then read it back (still as v1beta2), is the defaultEventList still there?

Anyway: before doing much more on this, have a read of https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md if you haven't already. I think there's a few interesting topics in there we should discuss tomorrow.

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
koshiksamota10: I am just getting rid of the conversionReviewVersions field as it is not necessarily needed by v1beta1 apiextension version.

joel-bluedata:
Closing this for now. Save the branch!

PR #314, opened 2020-05-21
ensure systemd is up before service enable/start

### LINKED ISSUE #315: ensure systemd is up before service enable/start
Seems that there can be a startup race before systemd services can be properly started inside nodes. Noticing this on K8s nodes that use docker-ce. So let's be cautious and not run startscripts until systemd is working (for apps that require it).

joel-bluedata:
fixes #315 

Seems that there can be a startup race before systemd services can be properly started inside nodes. Noticing this on K8s nodes that use docker-ce. So let's be cautious and not run startscripts until systemd is working (for apps that require it).

Also a fix in cluster.go here for something that golint was complaining about.

Should backport this to 0.4.2.

### DIFF
--- a/pkg/controller/kubedirectorcluster/members.go
+++ b/pkg/controller/kubedirectorcluster/members.go
@@ -33,6 +33,7 @@ import (
 	"github.com/bluek8s/kubedirector/pkg/shared"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/util/exec"
 )
 
 // syncMembers is responsible for adding or deleting members. It and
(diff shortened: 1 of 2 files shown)


pkg/controller/kubedirectorcluster/members.go
```diff
@@ -921,6 +922,48 @@ func generateNotifies(
...
+		podName,
+		expectedContainerID,
+		executor.AppContainerName,
+		"systemd check",
+		strings.NewReader(cmd),
+	)
+	if cmdErr == nil {
+		// No error, including in the return status of the command. All good.
+		return true, nil
+	}
```
swamibluedata, line 956: does it make sense to check for nativeSystemdSupport also?

swamibluedata: uh may be not

PR #316, opened 2020-05-22
set SpecGenerationToProcess earlier

### LINKED ISSUE #312: set SpecGenerationToProcess earlier
If we get an error in syncMemberServices (causing handler exit), we lose the necessary update to SpecGenerationToProcess.

We should set SpecGenerationToProcess as soon as we know we should. Right after syncClusterRoles should be good, since the "state" from that function is what we would be worried about losing an an error bailout. We don't need to worry about currentHash and cr.Status.LastConnectionHash since we'll get the same values for those on the next handler.

joel-bluedata:
fixes #312 

If we get an error in syncMemberServices (causing handler exit), we lose the necessary update to SpecGenerationToProcess.

Easiest way to trigger this is to run KD in an environment like GKE using LoadBalancer services that can take a while to delete. Create a cluster with a large number of role members, then edit the cluster to shrink that role to its minimum, then immediately edit the cluster again to size the role up again. Because the old services are still being deleted and we want to re-use the same service names, reconciliation will fail until the old services finish deleting.

We should set SpecGenerationToProcess as soon as we know we should. Right after syncClusterRoles should be good, since the "state" from that function is what we would be worried about losing an an error bailout. We don't need to worry about currentHash and cr.Status.LastConnectionHash since we'll get the same values for those on the next handler (unless currentHash legitimately changes).

With this change the above test scenario works, i.e. reconciliation proceeds after the old services go away,

### DIFF
--- a/pkg/controller/kubedirectorcluster/cluster.go
+++ b/pkg/controller/kubedirectorcluster/cluster.go
@@ -202,6 +202,17 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 		return rolesErr
 	}
 
+	// The "state" calculated above can be different on next handler pass,
+	// so we need to make sure we bump the spec gen now if necessary.
+	// If we delay doing this, a handler error (e.g. in syncMemberServices)
+	// could cause a handler exit and we would lose the necessary spec gen
+	// update.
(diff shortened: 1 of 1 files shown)


kmathur2:
makes sense.

PR #321, opened 2020-06-04
BDP 1319: Introduce a namingScheme for KD K8s objects.

koshiksamota10:
Introduce a naming scheme that can be configured to generate names for Kubernetes objects based on user's choice.


deploy/kubedirector/kubedirector.hpe.com_kubedirectorconfigs_crd.yaml
```diff
@@ -46,6 +46,8 @@ spec:
               type: string
             clusterSvcDomainBase:
               type: string
+            defaultNamingScheme:
+              type: boolean
```
joel-bluedata, line 50: I think we want to leave ourselves some wiggle room for future changes with this configurability... rather than make it a boolean switch, it should probably be an enum, handled similarly to defaultServiceType. (This applies to the kdcluster-level property too of course.)

As for what the allowed values of the enum should be... well, for now there are two. For "old style" maybe call it something like "UID" or "Generate"; for "new style" maybe "ClusterAndRole".

Also, in order to be backwards compatible, the default value for this top-level config property should be the one for old style, so that behavior is the same in new KD releases unless someone explicitly chooses otherwise.

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
```diff
@@ -96,8 +96,6 @@ spec:
                         type: string
                       isDashboard:
                         type: boolean
-                      hasAuthToken:
-                        type: boolean
```
joel-bluedata, line 100: Are these some accidental deletions from this CRD? Merge issues?

koshiksamota10: Yes.

koshiksamota10:
I have made the changes to support multiple naming schemes. For now, we have 2: ClusterRole which is the new naming scheme and UID which is the old one. The default is set to the old one.

I have tried some of the scenarios which include:
1) Defining no defaultNamingScheme or /cluster namingScheme.
2) Defining just the defaultNamingScheme.
3) Defining defaultNamingScheme and overriding it with /cluster naming scheme.
4) Defining just /cluster naming scheme.
5) Various clusters having different naming schemes.

I have set the max name limit to 56 which can obviously be modified.

Also, tested cases where namingScheme exceeds this threshold or we pass the wrong scheme name.

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
```diff
@@ -19,6 +19,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	// UID represents the old naming scheme where object names were genrated
```
joel-bluedata, line 23: typo genrated

pkg/validator/types.go
```diff
@@ -76,4 +76,7 @@ const (
 
 	invalidResource = "Specified resource(\"%s\") value(\"%s\") for role(\"%s\") is invalid. Minimum value must be \"%s\"."
 	invalidSrcURL   = "Unable to access the specified URL(\"%s\") in file injection spec for the role (%s). error: %s."
+
+	nameLimit       = "The KD cluster name violates the name length limit. The name is a combination of cluster name + role name."
```
joel-bluedata, line 80: Let's phrase this a little differently. Hmm. How about something like:

"When using the ClusterRole naming scheme, the total length of KD cluster name + any of its role IDs must not exceed 56 characters."

(And ideally we'd have a shared constant for that "56" amount that can be used both here as well as in validator/cluster.go, in case we need to change the limit later.)

joel-bluedata:
I think this looks really good! A couple of minor comments above.

One general quibble: there are a lot of other unrelated commits in this branch, although it looks like most have been reverted. I'd rather have a clean merge of only related commits. Let me know how you want to handle this... one approach would be to make a new branch off of master and cherrypick the related commits into there.

koshiksamota10:
> I think this looks really good! A couple of minor comments above.
> 
> One general quibble: there are a lot of other unrelated commits in this branch, although it looks like most have been reverted. I'd rather have a clean merge of only related commits. Let me know how you want to handle this... one approach would be to make a new branch off of master and cherrypick the related commits into there.
> 
> (I can handle that branch cleanup if you would rather not.)

Umm, my local master branch was messed up with all the crd conversion changes that I was doing and I thought we were going to push those changes. I had to revert all of them on master. When I checked out my new branch off master, it brought along all those commits and some other changes that I hadn't reverted. Let me figure out a way to clean this up and stabilize my master branch and name scheme branch.

Issue #317, opened 2020-05-26
don't include zero-member roles in the generated configmeta


Pandoraemon:
[root@kdss-hp69z-0 log]# python /opt/guestconfig/cdh6-cm-setup/setup_cluster.py
('output of NodeGroupMacro.getLocalGroupHosts()', [u'kdss-xjzd5-0.kdhs-vk48x.cdh.svc.cluster.local', u'kdss-b4g2k-0.kdhs-vk48x.cdh.svc.cluster.local', u'kdss-b4g2k-2.kdhs-vk48x.cdh.svc.cluster.local', u'kdss-hp69z-0.kdhs-vk48x.cdh.svc.cluster.local', u'kdss-szn5x-0.kdhs-vk48x.cdh.svc.cluster.local', None, u'kdss-b4g2k-1.kdhs-vk48x.cdh.svc.cluster.local'])
Traceback (most recent call last):
  File "/opt/guestconfig/cdh6-cm-setup/setup_cluster.py", line 45, in <module>
    HOST_INDEX_MAP[H] = NodeMacro.getNodeIndexFromFqdn(H)
  File "/usr/lib/python2.7/site-packages/configcli-1.0-py2.7.egg/configcli/macro/node.py", line 94, in getNodeIndexFromFqdn
    node_id = self.getNodeIdFromFqdn(fqdn)
  File "/usr/lib/python2.7/site-packages/configcli-1.0-py2.7.egg/configcli/macro/node.py", line 112, in getNodeIdFromFqdn
    raise Exception("Failed to get node Id for the given FQDN " + str(fqdn))
Exception: Failed to get node Id for the given FQDN None

Pandoraemon:
happend when I use CDH6

omhpe:
Hi @kcsraju ,

We are using this to find all the hosts in a cluster using configcli

macroObj = configcli.getCommandObject("macro")
NodeMacro = macroObj.getSubcommandObject("node")
NodeGroupMacro = macroObj.getSubcommandObject("nodegroup")
ALL_HOSTS = NodeGroupMacro.getLocalGroupHosts()

In one of the scenarios faced by @Pandoraemon , it seems one of the value is None in ALL_HOSTS

[...]

PR #328, opened 2020-06-08
Make changes to skip roles which do not have any members.

koshiksamota10:
https://github.com/bluek8s/kubedirector/issues/317


joel-bluedata:
Can we sanity-check some app deployments with this (making sure to have some zero-member roles in the kdcluster)... especially CDH6?

koshiksamota10:
> Can we sanity-check some app deployments with this (making sure to have some zero-member roles in the kdcluster)... especially CDH6?

I did. I created a cdh cluster with few roles having 0 members.

PR #334, opened 2020-06-12
MapR 6.1 MEP 6.3 KD app

abhinandantn:
MapR app on KD version 1.0
- This version of the app is designed for KD and also each services is made to run independent pods, while in EPIC few services were grouped
- There is no need for repo location because all the required services are already installed on docker file, based on pod respective services gets enabled and configured
- As per the requirement for 5.1, it was decided to have only MapR core functionality, Hive, Hue and Edge services.
- Through secrets user can change the default MapR admin password and also pass the ssl_truststore and mapr-clusters.conf information to the pod (currently provided as a workaround). Until this feature gets in KD.
- There are 2 dockerfiles one for MEP and other one for edge pod. The reason for having 2 is because few packages cannot co-exist, mapr-core and mapr-client packages cannot co-exist.

### DIFF
--- a/deploy/example_catalog/cr-app-mapr610-secured.json
+++ b/deploy/example_catalog/cr-app-mapr610-secured.json
@@ -1,0 +1,256 @@
+{
+  "apiVersion": "kubedirector.hpe.com/v1beta1",
+  "kind": "KubeDirectorApp",
+  "metadata": {
+    "name" : "mapr610"
+  },
+
+  "spec" : {
+    "systemdRequired": true,
+    "defaultPersistDirs" : ["/opt/mapr"],
+    "config": {
+      "configMeta" : {
+               "storage" : "60",
+               "secure":"true"
+      },
+      "roleServices": [
+        {
+                "roleID": "control-system",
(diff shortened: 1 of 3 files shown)


joel-bluedata:
Lots of indentation/format variation in the JSON... can you run it through a formatter to clean it up?

abhinandantn:
I have fixed the indentation of JSON and YAML files

deploy/example_catalog/cr-app-mapr610-secured.json
```diff
@@ -0,0 +1,340 @@
...
+        }
+      ],
+      "selectedRoles": [
+        "control-system",
+        "resource-manager",
+        "history-server",
+        "zookeeper",
+        "cldb",
+        "nodemanager",
+        "fileserver"
```
joel-bluedata, line 116: Until we introduce choice-driven config variations this probably won't matter much and will seem like redundant busywork, but... the selectedRoles here should list all of the defined roles in the app.

abhinandantn: i have updated the json file to include all roles under "selectedRoles"

PR #338, opened 2020-06-17
removed package version of appconfig scripts inside image

Supekar:


joel-bluedata:
This works with the current image tag in the kdapp?

BTW I would strongly recommend using something other than "latest" for your image, since this will prevent existing kdapp installations from breaking when the image later changes. Let me know if there's some issue with this.

kmathur2:
> This works with the current image tag in the kdapp?
> 
> BTW I would strongly recommend using something other than "latest" for your image, since this will prevent existing kdapp installations from breaking when the image later changes. Let me know if there's some issue with this.

@joel-bluedata  The reasoning for the 'latest' tag is to avoid making changes to KD as we find and fix issues in app config/docker scripts. Bad idea?

swamibluedata:
Also pl. remove docker.io prefix from all the json

joel-bluedata:
@kmathur2 OK I see. We do need to be double sure not to leave it in that state for any official release though. If you guys want to handle things that way I'm willing to try it but file a GitHub issue against the 0.5 milestone here, with a list of the specific kdapps that need to get final image tags.

Supekar:
maybe if the tag 'latest' isn't suitable, we can set it to something else for all images for the latest release

joel-bluedata:
Well, the general point is that when we release something the tag it refers to should always point to the same image and never change.

Anyway as I said above I'm fine with "latest" for now (during ongoing development), but if you do want to do that then file an issue to fix it before release.

Swami also had a concern above.

joel-bluedata:
Just FYI before we merge this I'm waiting on one of two things: either using a specific tag (not latest), or else an issue filed to do so before 0.5 release.

Supekar:
Created a Jira ticket for KD app labels to be updated.
https://jira-pro.its.hpecorp.net:8443/browse/BDP-1585

Issue #340, opened 2020-06-18
duplicate distro IDs in example catalog


joel-bluedata:
cr-app-mapr610-secured.json and cr-app-mapr610-unsecured.json use the same distro ID. This breaks "make deploy".

joel-bluedata:
Also they use just the same resource name.

PR #342, opened 2020-06-19
Updated distroID of unsecured KD MapR app

abhinandantn:
this change is against the issue
  https://github.com/bluek8s/kubedirector/issues/340

fixed by modifying json file of unsecured KD MapR app


riteshja:
@abhinandantn 
As discussed in our call, I think we should change the distro-id of the secure version to bluedata/mapr610-secure and leave the distro-id of unsecure version as is. 

Same thing for the metadata.name. Secure version can be called mapr610-secure and unsecure version can remain as mapr610.

PR #354, opened 2020-06-23
KD Issues 336 and 337

koshiksamota10:
Introduce a limit on the number of total members for a KD cluster.
Change the statusChange flag to false in case the cluster is being
deleted and update write back fails.

### DIFF
--- a/pkg/validator/types.go
+++ b/pkg/validator/types.go
@@ -80,5 +80,5 @@ const (
 	nameLimit       = "When using the ClusterRole naming scheme, the total length of KD cluster name + any of its role IDs must not exceed 56 characters."
 	badNamingScheme = "Naming scheme not valid in the config file."
 
-	maxMemberLimit = "Maximum number of total members per KD cluster supported is 1000"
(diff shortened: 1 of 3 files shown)


pkg/validator/types.go
```diff
@@ -79,4 +79,6 @@ const (
 
 	nameLimit       = "When using the ClusterRole naming scheme, the total length of KD cluster name + any of its role IDs must not exceed 56 characters."
 	badNamingScheme = "Naming scheme not valid in the config file."
+
+	maxMemberLimit = "Maximum number of total members per KD cluster supported is 1000"
```
joel-bluedata, line 83: One tweak. Let's make maxMemberLimit have a `%d` token instead of hardcoding 1000 there. Then the code in validateCardinality can use fmt.Sprintf to compose the actual error message by substituting in the value of maxKDMembers.

joel-bluedata:
All righty!

FYI I think I'm not going to make a special tagged KD build for this. I'll have another KD change following along behind & I'll sweep it up in that.

PR #359, opened 2020-06-29
Misc KD Fixes/gaps exposed during recent KDApp testing

kmathur2:
This PR has few minor fixes 

1.  If a role has members who do not have hasAuthToken enabled then authToken was being overwritten and was not getting populated in the configmeta JSON.

2. For connected secret and configmaps if multiple secret or configmaps have same cmType (deduced by the label) they were being overwritten.
The new structure would be map of resourceType: List, where a List is a list of secrets or configmaps.

3. Propagate kdapp labels to kdcluster and all kdcluster labels to service. This is good metadata to have which simplifies querying relevant resources.


pkg/shared/client.go
```diff
@@ -210,6 +210,18 @@ func Update(
...
+// are changing the object.
+func Patch(
+	ctx context.Context,
+	obj runtime.Object,
+	patchspec k8sClient.Patch,
+) error {
+
+	return client.Patch(ctx, obj, patchspec)
+}
+
```
joel-bluedata, line 224: Is this used anywhere?

Note that in HPECP Agent we have it defined as
```go
// Patch uses the split client. Should write back directly to K8s, but we'll
// use the split client in case it ever wants to use the knowledge that we
// are changing the object.
func Patch(
	ctx context.Context,
	originalObj runtime.Object,
	modifiedObj runtime.Object,
) error {

	patch := k8sClient.MergeFrom(originalObj)
	return client.Patch(ctx, modifiedObj, patch)
}
```

...so if/when we need it here in KD we should probably do the same to avoid confusion.

pkg/executor/util.go
```diff
@@ -127,6 +134,10 @@ func labelsForService(
 			result[name] = value
 		}
 	}
+	// Apply all cluster labels also
+	for name, val := range cr.Labels {
+		result[name] = val
+	}
```
joel-bluedata, line 140: I'm not truly opposed to this but it does raise some questions. E.g.
* Why apply these to just services (and not pods, pvcs, statefulsets, whatever)?
* Does this raise the expectation that these labels would track the kdcluster's labels (whereas we're actually just doing a one-time copy)?

Was there a specific usecase that this helps? (And something that couldn't be satisfied by first determining the kdcluster and then going to look at its labels.)

pkg/executor/util.go
```diff
@@ -68,6 +68,13 @@ func labelsForCluster(
 		ClusterAppLabel:        cr.Spec.AppID,
 		ClusterAppCatalogLabel: *(cr.Spec.AppCatalog),
 	}
+	appCR, err := catalog.GetApp(cr)
+	if err == nil {
+		// Apply app labels to cluster
+		for name, val := range appCR.Labels {
+			result[name] = val
+		}
+	}
```
joel-bluedata, line 77: Would we want to put the kdapp's labels on the cluster components (which is what this does) or would we rather put them on the kdcluster itself? Or both?

pkg/catalog/configmeta.go
```diff
@@ -177,20 +177,29 @@ func servicesForRole(
...
-	kdcm := make(map[string]map[string]map[string]string)
+	// Many connected configmaps can be of a given type, hence
+	// create a map of cmType:list of configmaps, where
+	// every configmap is a map of string and string
+	kdcm := make(map[string][]map[string]map[string]string)
 	for _, connectedCmName := range cr.Spec.Connections.ConfigMaps {
 		cm, err := observer.GetConfigMap(cr.Namespace, connectedCmName)
 		if kdConfigMapType, ok := cm.Labels[configMapType]; ok {
+			cmMap := make(map[string]map[string]string)
 			cmMap[connectedCmName] = cm.Data
```
joel-bluedata, line 195: So the map at this level will always only have one key? Hmm.

I mean this is fine if this is the structure that works best for the consumer, but it seems like semantically the slice elements are a pair of name-and-map, rather than a map of maps.

kmathur2: Ya, that is right, we look for a specific key for config map and secrets. We can add more key(s) in the future. Having said that this is already quite flexible. We can store all maps, secrets of a given type as a list, and it's intuitive for apps to query in those ways.

pkg/executor/util.go
kmathur2: good point, I don't feel strongly one way or the other. I just used the convenience fun that puts crName, appid etc

pkg/executor/util.go
kmathur2: Yes, there is a specific use case in MLOps tenant since we branch clusters into Notebook, Deployment, and Training and we would like to segregate service Endpoints accordingly. That is where services are different than other resources like Pod, statefulset, etc since services are directly consumed by Enduser and user needs to know more about them, the same is not true for Pods, stateful sets, deployments, etc. If need be we can continue to propagate on other resources as well.

kmathur2: Apparently the current UI design makes it hard to lookup by determining the kdcluster and then going to look at its labels. @Nabin and @Rajus can comment more on this

pkg/shared/client.go
kmathur2: Yes, I was planning to use a patch for annotations update. Will try this and change accordingly.

pkg/catalog/configmeta.go
joel-bluedata: We might not be talking about the same thing so let me get into some details...

1. We have a top-level map where key is "type", and value is a list.
2. In the list, each element is a map.
3. In this map, there is always only one entry. The key is the name of the connected resource, and the value is its data (which is another map).

It's level 3 that seems a little odd. A map that always only has one entry is really just a pair, which might more naturally represented as a list-of-2-things or a struct-with-2-fields.

Are you saying that that map at level 3 could have different keys in the future?

kmathur2: ok, yes you are spot on. The 3rd level key is the name of the resource. We could put more stuff there like labels, annotations since only data is being propagated now. Does it make sense?

pkg/executor/util.go
joel-bluedata: Got it. OK, let me summarize my understanding of the current label-related changes here.

* Whenever any statefulset is created by us, we copy to it all current labels on the kdapp. This also includes setting the labels in the statefulset pod templates, which the statefulset will later use to create pods.

* Whenever any service is created by us, we copy to it all current labels on the kdapp and all current labels on the kdcluster.

This feels a little shaky to me. For example let's say that a kdapp has some label foo=bar and we create a kdcluster from it that has 1 role with 1 member. The kdcluster has the label name=joel.

We'll end up with:
```
headless service labels foo=bar, name=joel
statefulset 0    labels foo=bar, name=joel
pod 0-0          labels foo=bar, name=joel
service 0-0      labels foo=bar, name=joel
```
Now we edit the kdapp label to be foo=zowie, and then we expand the role to 2 members.
```
headless service labels foo=bar, name=joel
statefulset 0    labels foo=bar, name=joel
pod 0-0          labels foo=bar, name=joel
service 0-0      labels foo=bar, name=joel
pod 0-1          labels foo=bar, name=joel
service 0-1      labels foo=zowie, name=joel
```
Now we edit the kdcluster label to be name=kartik, and we add a member in another role.
```
headless service labels foo=bar, name=joel
statefulset 0    labels foo=bar, name=joel
pod 0-0          labels foo=bar, name=joel
service 0-0      labels foo=bar, name=joel
pod 0-1          labels foo=bar, name=joel
service 0-1      labels foo=zowie, name=joel
statefulset 1    labels foo=zowie, name=kartik
pod 1-0          labels foo=zowie, name=kartik
service 1-0      labels foo=zowie, name=kartik
```
If at this point we edited the kdapp label back to foo=bar and expanded the second role, we'd have elements of this one kdcluster with all four possible combinations of those label values.

I think some label propagation from the kdapp to the kdcluster and its components would be good. But we have to be careful about how & when we snapshot them, unless we're going to commit to tracking them for reconciliation.

I don't think that it's sufficient to assume that the labels will never be edited. And we can't choose to enforce that either, unless we limit ourselves to a very particular label key.

Hmm.

==========

To be sure I understand the usecase before proceeding:

We have labels on the kdapps, and we want to get those labels onto the services.

Is there no need to get those labels onto the kdcluster resources too?

Is it just a very specific "app category" label that we care about?

pkg/catalog/configmeta.go
joel-bluedata: Got it, thanks!

joel-bluedata: Although what would the keys look like then? Currently the "lone key" could be anything. The resource itself could even be named "labels" in which case the key would be "labels".

If we did add other things to this map like labels and annotations what would the map then look like?

pkg/executor/util.go
kmathur2: hmm, actually the current requirement is just to get the kdcluster labels to service. That should be ok. Currently, the UI is applying some labels on kdcluster which I thought could potentially be avoided by propagating all the way from app to cluster to services. If I look at above example then you are right it can get a bit hairy. So it's getting unnecessarily complicated for now so  I will leave it for now

kmathur2: The label we care about right now is createdByUser which UI is adding on kdcluster. Based on that the UI will display only the Notebook cluster and service endpoints for the logged-in user.

joel-bluedata: The UI could explicitly put that in the serviceLabels in the role config of the kdcluster...

pkg/catalog/configmeta.go
kmathur2: Currently, the DS looks like
```
"security" // cmType : [  {"ldapinfo" :  
                                           {<k1 : v1>, <k2 : v2>}},
                                          {"kerberos" : 
                                            {<k1 : v1>, <k2 : v2>}}
                                     ] 
```
When we propagate labels and keys it should be something like -

```
"security" // cmType : [  {"ldapinfo/data" :  
                                           {<k1 : v1>, <k2 : v2>},
                                        {"ldapinfo/labels" : 
                                           {<k1 : v1>, <k2 : v2>},
                                        {"ldapinfo/annotations" : 
                                           {<k1 : v1>, <k2 : v2>}, 
                                       // and more metadata about the resource

                                     ] 
```

pkg/executor/util.go
kmathur2: hmmm, that should be ok too, so you are suggesting lets table the label propagation for now and revisit later since labels can change?

joel-bluedata: yup, if possible

kmathur2: ok, I will work with UI folks. We will come back to this if we hit a stalemate.

pkg/catalog/configmeta.go
joel-bluedata: Do you mean this:
```
"security" // cmType : [{"ldapinfo/data" : {<k1 : v1>, <k2 : v2>},
                         "ldapinfo/labels" : {<k1 : v1>, <k2 : v2>},
                         "ldapinfo/annotations" : {<k1 : v1>, <k2 : v2>},
                         // etc.
                        },

                        {"kerberos/data" : {<k1 : v1>, <k2 : v2>},
                         // etc.
                        }] 
```
I.e. the ldap/\* stuff are all keys in one of the 3rd-level maps.

I'm not 100% clear I guess on how the consumer is supposed to use this. Do they typically need to know the resource name? Will they be walking the keys and splitting each key on slash to figure out what the value is representing?

Maybe it should be something like
```
"security" // cmType : [{"metadata" : {<k1 : v1>, <k2 : v2>},
                         "data" : {<k1 : v1>, <k2 : v2>},
                         "labels" : {<k1 : v1>, <k2 : v2>},
                         "annotations" : {<k1 : v1>, <k2 : v2>}}]
```
where metadata has at least the resource name and could also carry kind and namespace for the heck of it. For this release we could populate the metadata and data maps.

But you know more than I do about how this is currently being used.

In any case I think we should establish the scheme now rather than planning to immediately change it in the next release, even if we're not yet putting labels or annotations in there.

kmathur2: okay, yea, I overengineered in my example but you got the idea. Will adhere to your example structure

kmathur2:
@joel-bluedata  Done with changes

joel-bluedata:
I'll do a tagged image build when https://github.com/bluek8s/kubedirector/pull/361 is merged. Expecting tonight or tomorrow for that ... let me know if you need one before then.

kmathur2:
Tomorrow is ok 👍

PR #360, opened 2020-06-30
BDP Issue 353

koshiksamota10:
If the storage stanza is specified, the size must be > 0. Otherwise the statefulset controller will freak out.

### DIFF
--- a/pkg/validator/cluster.go
+++ b/pkg/validator/cluster.go
@@ -437,7 +437,7 @@ func validateRoleStorageClass(
 			)
 			break
 		}
-		if storageSize.IsZero() {
+		if storageSize.Sign() == 1 {
 			valErrors = append(
 				valErrors,
 				fmt.Sprintf(
--- a/pkg/validator/cluster.go
+++ b/pkg/validator/cluster.go
@@ -435,7 +435,7 @@ func validateRoleStorageClass(
 					role.Name,
 				),
 			)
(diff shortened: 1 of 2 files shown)


pkg/validator/cluster.go
```diff
@@ -424,6 +425,28 @@ func validateRoleStorageClass(
...
+			valErrors = append(
+				valErrors,
+				fmt.Sprintf(
+					invalidStorageDef,
+					role.Name,
+				),
+			)
+			break
+		}
+		if storageSize.IsZero() {
```
joel-bluedata, line 440: Can we check for a positive value here rather than just nonzero?

E.g. maybe `storageSize.Sign() == 1`

pkg/validator/cluster.go
```diff
@@ -424,6 +425,28 @@ func validateRoleStorageClass(
...
+			valErrors = append(
+				valErrors,
+				fmt.Sprintf(
+					invalidStorageDef,
+					role.Name,
+				),
+			)
+			break
+		}
+		if storageSize.Sign() == 1 {
```
joel-bluedata, line 440: this should be `!=` ?

koshiksamota10: How stupid of me :( Changed it.

joel-bluedata:
I'll roll these changes up into a later image build.

Issue #319, opened 2020-05-28
allow PUT to an in-use kdapp if nothing is being changed


joel-bluedata:
Currently the validator disallows PUT to a kdapp if a kdcluster is referencing it.

Arguably we should allow the PUT if it is not going to change the spec. Some automation usecases could benefit from this.

It's fairly simple to allow the PUT if the incoming spec is exactly the same as the spec in the stored resource; however I think we will first need to handle population-of-defaults before comparing against the stored spec.

PR #361, opened 2020-06-30
allow PUT to an in-use kdapp if nothing in the spec is being changed

joel-bluedata:
closes issue #319

It's not super-trivial to see whether an incoming spec changes anything, because of the way we handle global defaults and apply them to roles. So the validateRoles function here is changed to not only generate patches to modify the resulting document, but also apply the same changes to the in-memory appCR so we can use it for comparisons if this is a PUT.

Then admitAppCR does the comparison.

There's a big comment added to admitAppCR so more about that below...

==========

This change also fixes a minor bug in our defaults handling. The pattern is that we apply the global default to the roles (if necessary) and then remove the default from the spec. However in the case of defaultConfigPackage, if it was explicitly set to null, we were not removing it.

Among our own example apps, this only happened in our "vanilla OS" apps (for CentOS etc.) and is functionally harmless, but I needed to fix this behavior so that everything worked uniformly and I could do the necessary comparisons.

That does mean that any previously-created kdapps with a null defaultConfigPackage would not be able to take advantage of this change. That exception might kind of mess with upgrade processes. So I made the additional change in admitAppCR to ignore any difference in defaultConfigPackage value. These defaults don't directly affect how kdclusters behave in any case; they only affect the roles (which we ARE checking for differences).

### DIFF
--- a/pkg/validator/app.go
+++ b/pkg/validator/app.go
@@ -17,6 +17,7 @@ package validator
 import (
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"strconv"
 	"strings"
 
@@ -151,19 +152,26 @@ func validateSelectedRoles(
 // validateRoles checks each role for property constraints not expressible
 // in the schema. If any overrideable properties are unspecified, the corresponding
 // global values are used. This will add an PATCH spec for mutation the app CR.
+// The role in appCR will be correspondingly updated so that it can later be
+// to check whether the resulting CR differs from the current stored appCR.
 func validateRoles(
 	appCR *kdv1.KubeDirectorApp,
 	patches []appPatchSpec,
 	valErrors []string,
 ) ([]appPatchSpec, []string) {
 
+	// Any global defaults will be removed from the CR. Remember their values
+	// though for use in populating the role definitions.
 	var globalImageRepoTag *string
 	var globalSetupPackageURL *string
 	var globalPersistDirs *[]string
 	var globalEventList *[]string
-
-	globalImageRepoTag = appCR.Spec.DefaultImageRepoTag
-	if globalImageRepoTag != nil {
+	if appCR.Spec.DefaultImageRepoTag == nil {
+		globalImageRepoTag = nil
+	} else {
+		tagCopy := *appCR.Spec.DefaultImageRepoTag
+		globalImageRepoTag = &tagCopy
+		appCR.Spec.DefaultImageRepoTag = nil
 		patches = append(
 			patches,
 			appPatchSpec{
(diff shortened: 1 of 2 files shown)


pkg/validator/app.go
```diff
@@ -151,19 +152,26 @@ func validateSelectedRoles(
...
 	patches []appPatchSpec,
 	valErrors []string,
 ) ([]appPatchSpec, []string) {
 
+	// Any global defaults will be removed from the CR. Remember their values
+	// though for use in populating the role definitions.
 	var globalImageRepoTag *string
 	var globalSetupPackageURL *string
 	var globalPersistDirs *[]string
 	var globalEventList *[]string
```
koshiksamota10, line 168: I would leave a blank line here for more readability but that is how I like to read my code :)

pkg/validator/app.go
```diff
@@ -172,12 +180,16 @@ func validateRoles(
 			},
 		)
 	}
-
-	if (appCR.Spec.DefaultSetupPackage.IsSet == false) || (appCR.Spec.DefaultSetupPackage.IsNull == true) {
+	if !appCR.Spec.DefaultSetupPackage.IsSet {
```
koshiksamota10, line 183: Again nitpicking but can we reverse the blocks here? Like just have appCr.Spec.DefaultSetupPackage.IsSet() and then if it is not set initialize it to nil?

joel-bluedata: Sticking to the pattern of the other blocks below I think, unless you'd prefer those be flipped too.

In any case lemme wait and see if there are any other review code changes that would require re-testing.

pkg/validator/app.go
```diff
@@ -151,19 +152,26 @@ func validateSelectedRoles(
 // validateRoles checks each role for property constraints not expressible
 // in the schema. If any overrideable properties are unspecified, the corresponding
 // global values are used. This will add an PATCH spec for mutation the app CR.
+// The role in appCR will be correspondingly updated so that it can later be
+// to check whether the resulting CR differs from the current stored appCR.
```
kmathur2, line 156: nit: later be used i think

pkg/validator/app.go
kmathur2: globalDefaults here means common across all roles, is it?

joel-bluedata: yep

kmathur2:
General question, we want to block if the spec is being changed if cluster is running. So are we doing this to allow changes to label/annotations? What about the name ?

joel-bluedata:
Anything outside the spec is OK with us.

K8s itself will block changes to metadata like name/kind/namespace.

This change allows changes to the spec that "don't matter". You should be able to PUT the resource over and over as long as the resulting document (after defaults substitutions) is the same. This allows easier automation of processes that work with kdapps.

kmathur2:
> Anything outside the spec is OK with us, so labels and annotations can change whenever. This has always been the case.
> 
> K8s itself will block changes to metadata like name/kind/namespace.
> 
> This change allows changes to the spec that "don't matter". You should be able to PUT the resource over and over as long as the resulting document (after defaults substitutions) is the same. This allows easier automation of processes that work with kdapps.

ok, so before even if you apply the same YAML we would have not allowed and with these changes, the scripts won't complain. Was there a specific use case or is it just to provide standard k8s behavior?

joel-bluedata:
Yeah we can follow up more in email but this is related to a KD trial at a company.

kmathur2:
ok looks good 👍

PR #365, opened 2020-07-06
removing docker.io from JSON file

abhinandantn:
2 JSON file had docker.io and so removed them.


joel-bluedata:
From the KD point of view this isn't necessary, but I understand you have other reasons for scrubbing that.

PR #368, opened 2020-07-10
Cluster Connection deletion fix

kmathur2:
This PR tackles the deletion of the connected cluster, because of which KD was panicking into an unrecoverable state.
Example:

Cluster A's connection
```
connections:
  clusters:
  - "clusterB"
``` 
If clusterB is deleted then KD was panicking while trying to calculate clusterB's configmeta.

Added the required checks to avoid such scenarios.

### DIFF
--- a/pkg/catalog/configmeta.go
+++ b/pkg/catalog/configmeta.go
@@ -28,7 +28,6 @@ import (
 	"github.com/bluek8s/kubedirector/pkg/shared"
 	"github.com/google/uuid"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 )
 
 const (
@@ -190,6 +189,9 @@ func genConfigConnections(
 	kdcm := make(map[string][]map[string]map[string]string)
 	for _, connectedCmName := range cr.Spec.Connections.ConfigMaps {
 		cm, err := observer.GetConfigMap(cr.Namespace, connectedCmName)
+		if err != nil {
+			continue
+		}
 		if kdConfigMapType, ok := cm.Labels[configMapType]; ok {
 			cmMap := make(map[string]map[string]string)
 			cmMap["metadata"] = map[string]string{"name": cm.Name}
@@ -201,10 +203,6 @@ func genConfigConnections(
 			} else {
 				typeMaps := make([]map[string]map[string]string, 0)
 				kdcm[kdConfigMapType] = append(typeMaps, cmMap)
-			}
-
-			if err != nil {
-				return nil, err
 			}
 		}
 	}
@@ -224,6 +222,9 @@ func genSecretConnections(
 	kdsecret := make(map[string][]map[string]map[string][]byte)
 	for _, connectedsecretName := range cr.Spec.Connections.Secrets {
 		sec, err := observer.GetSecret(cr.Namespace, connectedsecretName)
+		if err != nil {
+			continue
+		}
 		if kdSecretType, ok := sec.Labels[secretType]; ok {
 			xlateMap := func(valueMap map[string]string) map[string][]byte {
 				convMap := make(map[string][]byte)
@@ -243,10 +244,6 @@ func genSecretConnections(
 				typeSecrets := make([]map[string]map[string][]byte, 0)
 				kdsecret[kdSecretType] = append(typeSecrets, secretMap)
 			}
-
-			if err != nil {
-				return nil, err
-			}
 		}
 	}
 	return kdsecret, nil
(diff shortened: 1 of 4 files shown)


pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -343,31 +343,37 @@ func calcConnectionsHash(
...
+		if clusterErr == nil {
+			// extra careful while dereferencing
+			if clusterObj.Status.SpecGenerationToProcess == nil {
+				specNum = "nil"
+			} else {
+				specNum = strconv.Itoa(
+					int(*clusterObj.Status.SpecGenerationToProcess))
+			}
 		}
 		buffer.WriteString(specNum)
```
joel-bluedata, line 358: If clusterErr is non-nil and therefore `string` is "" (not "nil"), does this code do what we want it to?

pkg/catalog/configmeta.go
```diff
@@ -384,13 +385,14 @@ func clusterBaseConfig(
 	kdConfigMaps, cmErr := genConfigConnections(cr)
 	kdSecrets, secErr := genSecretConnections(cr)
 
-	if cmErr != nil {
+	if !errors.IsNotFound(cmErr) && cmErr != nil {
```
joel-bluedata, line 388: This looks like it will bail out of gathering all connections info if any of the connected objects are missing. Am I reading that right?

kmathur2: If connection obj is missing then we keep going but if it's present BUT non-retrievable then we return. 
Hmmm now thinking again, I think we should never return, the values would be correctly set to nil if things were bad.

joel-bluedata: I'm also looking inside the functions like genClusterConnections -- looping over objects but immediately returning out of the loop if any error.

pkg/controller/kubedirectorcluster/cluster.go
kmathur2: yes I think so, if the cluster is non-nil then specNum would be "" which would generate new Hash and retrigger config meta generator which would reset cluster connections to nil inside pods which is what we want

pkg/catalog/configmeta.go
kmathur2: yea, so our strategy should be if any obj is not retrievable then ignore and keep going? For all obj (cluster, cm, and secrets )

joel-bluedata: I think so yes, especially in the not-found case. Skip the missing ones and just populate with everything else.

In the case where we actually get an API error (other than not-found) trying to retrieve something, that's arguably different. In that case I think we should probably not update any connections info at all on this handler pass and just try again next pass.

pkg/catalog/configmeta.go
```diff
@@ -189,6 +189,9 @@ func genConfigConnections(
 	kdcm := make(map[string][]map[string]map[string]string)
 	for _, connectedCmName := range cr.Spec.Connections.ConfigMaps {
 		cm, err := observer.GetConfigMap(cr.Namespace, connectedCmName)
+		if err != nil {
```
joel-bluedata, line 192: So e.g. going by my comment about error handling...

This seems good if the error is "not found".

However if the error is NOT "not found", but rather some other problem talking to the K8s API, we can return an error here. Let's see what would happen then...

* In the caller clusterBaseConfig we can return the error if any of the connection-generation functions return an error.
* In the caller of THAT, ConfigmetaGenerator, we can return the error if clusterBaseConfig returns an error.
* The caller of THAT, syncCluster, already handles those errors and exits the handler. We've already bumped SpecGenerationToProcess by that point, so we should be set to try again on the next handler (but please double-check that logic).

pkg/controller/configmap/configmap.go
```diff
@@ -84,6 +84,10 @@ func (r *ReconcileConfigMap) syncConfigMap(
 			for {
 				updateMetaGenerator := &kubecluster
 				annotations := updateMetaGenerator.Annotations
+				// if annotations == nil {
+				// 	annotations := make(map[string]string)
+				// 	updateMetaGenerator.Annotations = annotations
+				// }
```
joel-bluedata, line 90: What's up here?

pkg/controller/kubedirectorcluster/cluster.go
joel-bluedata: Hmm. FWIW this is different from how the configmap and secret connections are handled below; in those cases if the err is non-nil we don't write anything into the buffer.

pkg/controller/configmap/configmap.go
kmathur2: I am still trying to reproduce an issue I saw just once, my guess is - while the cluster is being deleted the cluster obj is there but annotation has been removed if that happens it might panic below. Not sure. Do you see a race window here?

joel-bluedata: Eh regardless, this block of code should not assume & depend on annotations being non-nil. It's good to have the check before we start trying to deref it.

joel-bluedata:
It all looks pretty good I think. Is it generally re-tested and ready to merge?

kmathur2:
> It all looks pretty good I think. Is it generally re-tested and ready to merge?

Trying one last bit, will update

kmathur2:
@joel-bluedata Happy with testing, this can be merged now

joel-bluedata:
I've pushed a tagged image at bluek8s/kubedirector:2703b10 if you want to use that.

kmathur2:
sounds good. Will update everest

PR #378, opened 2020-07-16
Updating MapR docker version in JSON file and added document for the App

abhinandantn:
Hi,
    KD MapR app has been updated to have generic license and so the dockerfile version was bumped. So had to change the JSON file to match the latest docker image version.

Also added a document on what features are added in the app and how to use the app with some sample programs.


joel-bluedata:
Thanks!

PR #388, opened 2020-07-22
don't use CR name in volume name

### LINKED ISSUE #387: possible to specify kdcluster name that prevents member pod creation
The kdcluster name is incorporated into the volume names for cgroup and cgroup-systemd, for kdclusters from systemd-using apps (cf. generateSystemdSupport). This can break things because some characters e.g. "." are valid in CR names but not in volume names. The symptom is that the statefulset will refuse to ever create the pods.

I don't immediately know of a reason why we need to incorporate the CR names into those volume names, so this might be a quick fix.

joel-bluedata:
Might have characters in it that are invalid for volume name.

closes #387 

@swamibluedata any possible issue with this?

### DIFF
--- a/pkg/executor/statefulset.go
+++ b/pkg/executor/statefulset.go
@@ -602,8 +602,8 @@ func generateSystemdSupport(
 	cr *kdv1.KubeDirectorCluster,
 ) ([]v1.VolumeMount, []v1.Volume) {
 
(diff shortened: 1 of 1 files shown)


swamibluedata:
So PVC names can have ".", but just not volumenames in statefulset?

PR #392, opened 2020-08-05
Create cr-app-gitlab-ee.json

hoycdanny:


deploy/example_catalog/cr-app-gitlab-ee.json
```diff
@@ -0,0 +1,65 @@
...
+            ],
+            "selectedRoles": [
+                "gitlab"
+            ]
+        },
+        "label": {
+            "name": "GitLab",
+            "description": "GitLab apps"
+        },
+        "distroID": "bluedata/gitlab",
```
joel-bluedata, line 29: Folks should start using some string other than "bluedata" for the vendor part of these distro IDs. Would you like to be the pioneer? :-)

hoycdanny: Ok, I will change distroID create new pull requests.

Issue #309, opened 2020-05-12
deterministic names for created services (and other objects?)


joel-bluedata:
The specific usecase is for Ingress backend, which can only be specified by service resource name.

One reason that current names are generated the way they are is to remove the possibility of collision, which is a good invariant to preserve. The resources related to individual members are based on the statefulset name, which is of course unique in the namespace. Cf. https://github.com/joel-bluedata/kubedirector/commit/4a577b4b7f9a9cff7e57453f03feae196175c70c

Secondarily we also keep the names short so they won't violate length restrictions.

We could introduce a new option to choose a naming style. Probably this would be an option in the KD config, rather than something chosen per-kdcluster? We don't want to allow free choice of names, but the names could be generated by different rules.

The obvious thing would be to base the names of objects on the kdcluster name and (if relevant) the role, but we would need to figure out how to handle length restrictions. As the obvious example: if the kdcluster name is already max length then you can't add prefixes/suffixes to it to create other names. Maybe the kdcluster validator could enforce more stringent restrictions on name length if this naming style is active.

In terms of addressing the Ingress usecase mentioned above, this naming method would make a service name entirely predictable given the kdcluster name.

joel-bluedata:
I think we're going to need to disable this feature in the 0.5.0 release of KD, and bake it some more.

There's a couple of easily correctable issues:
* The advertised PVC name is incorrect. The actual PVC name is "p-<crname>-<roleid>-<index>" but the member status says it is "<crname>-<roleid>".
* In retrospect the "ClusterRole" enum is probably unfortunate since "ClusterRole" is also a native K8s resource type; we should call that choice something else.

A larger issue though is that it is quite possible to have a kdapp/kdcluster combo that results in illegal object names with the current scheme.

The deterministic naming scheme in its current state is forming resource names as follows:
* Headless service => "<crname>"
* Statefulset => "<crname>-<roleid>"
* Member service is same as member pod name. Member pod name will be derived from statefulset => "<crname>-<roleid>-<index>".

All three of these can be invalid names.

The validation regexes for CR name and role ID are as follows:
* CR name => `[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)`
* kdapp role ID => `(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?`

And the regexes we have to fit into for a service or statefulset name:
[...]

PR #393, opened 2020-08-12
CrName + Role Naming Scheme

koshiksamota10:
Addressing Issue: https://github.com/bluek8s/kubedirector/issues/309

The new address scheme introduces a way to name the KD K8s objects in a more predictable and easily readable manner. The objects that are addressed by the changes include the pod name, statefulset name, pvc name and service name which are all a combination of Cr Name + Role name and a unique extension in some cases. 

In the straightforward case without munging the combinations follow the following pattern:
pvc name: p-cr name-role name-unique identifier-index string,
statefulset name: cr name-role name-unique identifier,
service name: cr name-role name-unique identifier-index string,
pod name: cr name-role name-unique identifier-index string.

This should match the description in kdcluster status.

The changes also include a validation scheme for the cases where the combination turns out to be longer than name length limit (46) in which case we ignore the characters beyond the limit. The uniqueness issue is solved by the unique extension generated by K8s.


joel-bluedata:
Plz add some PR description here about how the new scheme generates names for the various components, any validation and failure-case issues, or whatever else might be interesting/useful for docs and for code review.

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
```diff
@@ -19,6 +19,16 @@ import (
...
 )
 
+const (
+	// UID represents the old naming scheme where object names were generated
+	// with unique UID extensions.
+	UID string = "UID"
+
+	// CrNameRole represents the new naming scheme based on cluster name and
+	// respective role name.
+	CrNameRole string = "CrNameRole"
```
kmathur2, line 29: To utilize the new naming, will they have to put this string "CrNameRole" in their spec ?

koshiksamota10: Yes!!

pkg/executor/statefulset.go
```diff
@@ -273,18 +274,35 @@ func getStatefulset(
...
-		},
-		ObjectMeta: metav1.ObjectMeta{
+	namingScheme := *cr.Spec.NamingScheme
+	var objectName string
+	var objectMeta metav1.ObjectMeta
+	if namingScheme == v1beta1.CrNameRole {
+		objectName = ValidateName(cr.Name + "-" + role.Name)
+		objectName += "-"
+		objectMeta = metav1.ObjectMeta{
+			GenerateName:    objectName,
```
joel-bluedata, line 284: What's the need for using GenerateName in this scheme, as opposed to just Name?

koshiksamota10: You are correct, I dont think we need to use generate names here.

pkg/validator/cluster.go
```diff
@@ -807,6 +807,49 @@ func addServiceType(
...
+			clusterPatchSpec{
+				Op:   "add",
+				Path: "/spec/namingScheme",
+				Value: clusterPatchValue{
+					ValueStr: cr.Spec.NamingScheme,
+				},
+			},
+		)
+	} else {
+		if *cr.Spec.NamingScheme != kdv1.CrNameRole && *cr.Spec.NamingScheme != kdv1.UID {
```
joel-bluedata, line 840: Our schema validation should make this impossible.

pkg/executor/util.go
```diff
@@ -142,3 +144,28 @@ func createPortNameForService(
 	}
 	return strings.ToLower(portInfo.URLScheme) + "-" + portInfo.ID
 }
+
+// ValidateName is a utility function that limits object names to be below
+// nameLengthLimit threshold for the CrNameRole naming scheme.
```
joel-bluedata, line 149: Comment here doesn't fully describe what is going on in this function.

pkg/executor/service.go
```diff
@@ -62,8 +63,14 @@ func CreateHeadlessService(
 			},
 		},
 	}
+
+	namingScheme := *cr.Spec.NamingScheme
 	if cr.Status.ClusterService == "" {
-		service.ObjectMeta.GenerateName = headlessSvcNamePrefix
+		if namingScheme == v1beta1.CrNameRole {
+			service.ObjectMeta.Name = cr.Name
```
joel-bluedata, line 70: This isn't guaranteed to work. A kdcluster CR name can be of the form `[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)` while a service name must be of the form `[a-z]([-a-z0-9]*[a-z0-9])?`

joel-bluedata: (I think the difference there is just the dots.)

joel-bluedata: If the solution e.g. is just to remove the dots, then we might have to go back to using GenerateName here to guarantee uniqueness.

pkg/executor/statefulset.go
```diff
@@ -273,18 +274,35 @@ func getStatefulset(
...
-			Kind:       "StatefulSet",
-			APIVersion: "apps/v1",
-		},
-		ObjectMeta: metav1.ObjectMeta{
-			GenerateName:    statefulSetNamePrefix,
+	namingScheme := *cr.Spec.NamingScheme
+	var objectmeta metav1.ObjectMeta
+	var objectName string
+	if namingScheme == v1beta1.CrNameRole {
+		objectName = ValidateName(cr.Name + "-" + role.Name)
```
joel-bluedata, line 281: Similar comment here about uniqueness. Since ValidateName (probably should be called something else) will munge the name, it could result in the same name as some other existing statefulset. Actually I guess that's technically true even if we were not munging the name.

I know I had previously asked why there was any need to still use GenerateName in this scheme, so I guess that's the answer as to why.

pkg/validator/types.go
```diff
@@ -81,4 +81,7 @@ const (
 	invalidSrcURL   = "Unable to access the specified URL(\"%s\") in file injection spec for the role (%s). error: %s."
 
 	maxMemberLimit = "Maximum number of total members per KD cluster supported is %d."
+
+	nameLimit       = "When using the ClusterRole naming scheme, the total length of KD cluster name + any of its role IDs must not exceed 56 characters."
+	badNamingScheme = "Naming scheme not valid in the config file."
```
joel-bluedata, line 86: Can get rid of badNamingScheme if we remove the check for it as per the other comment.

pkg/validator/types.go
```diff
@@ -81,4 +81,7 @@ const (
 	invalidSrcURL   = "Unable to access the specified URL(\"%s\") in file injection spec for the role (%s). error: %s."
 
 	maxMemberLimit = "Maximum number of total members per KD cluster supported is %d."
+
+	nameLimit       = "When using the ClusterRole naming scheme, the total length of KD cluster name + any of its role IDs must not exceed 56 characters."
```
joel-bluedata, line 85: I don't think nameLimit is used anywhere anymore.

pkg/validator/cluster.go
```diff
@@ -807,6 +807,49 @@ func addServiceType(
...
 
+// validateNamingScheme function validates the namingScheme for all K8s objects.
+// If the naming scheme is unspecified, check to see if the default namingScheme
+// is provided through kubedirector's config CR, otherwise use a global constant
+// for naming scheme. In either of those cases add an entry to PATCH spec for mutating
+// cluster CR.
+// If the naming scheme is defined, validate the naming scheme for correctness
+// and that the cluster name, role name combination does not exceed the default object
+// length limit for K8s objects (63). The max length limit is set to 56 to account for
+// the additional '-' and/or digit characters that might be included in the name.
```
joel-bluedata, line 818: This second para of the header comment no longer seems relevant.

pkg/executor/util.go
```diff
@@ -24,6 +24,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
+const nameLengthLimit = 52
```
joel-bluedata, line 27: We probably need comments that have a detailed breakdown of how this limit is reached. I did some noodling myself to see if I could figure it out:

Max statefulset/pod/pvc name size is 253. Max service name size is 63, so that will be our bottleneck.

We use "crname-roleid" as the base prefix for creating the statefulset name, which then leads to the other names.

Let's say we're using GenerateName for the statefulset name (as per the other comments). GenerateName suffix adds 5 (usually/atleast) ... will eat into end of name if necessary, but let's try to avoid that. We'll probably want to add a dash at the end of our prefix in order to nicely separate it from the generated UID, so the whole cost of GenerateName would be 6 characters.

The service name will be statefulset-name plus dash plus index number. The index number could be up to 4 digits currently (cf. maxKDMembers). So the cost of those suffixes on the service name is 5 characters.

Therefore the length of the "crname-roleid" prefix + 5 + 6 must be <= 63. I.e. the length of the prefix must be <= 52.

Since this does depend on the number of digits in maxKDMembers, we probably want to either assert at runtime that maxKDMembers has no more than 4 digits, or else calculate the number of digits on the fly when creating this limit. If/when maxKDMembers becomes configurable that will be even more important to get right.

pkg/executor/util.go
```diff
@@ -142,3 +144,28 @@ func createPortNameForService(
...
+	name string,
+) string {
+	length := len(name)
+	var modName string
+
+	if length == 0 {
+		return name
+	}
+
+	for i := 0; i < length && i < nameLengthLimit; i++ {
```
joel-bluedata, line 160: If the limit is calculated as in my other comment then this loop could do "<= nameLengthLimit" instead of "< nameLengthLimit".

pkg/executor/statefulset.go
joel-bluedata: There's also the outside possibility that we would could munge two roles in the same kdcluster to have the same statefulset name. Assuming that we're using GenerateName then they will at least get names with different UID suffixes, so things will work, but it could be confusing.

However I don't think that's worth worrying much about. If someone e.g. names two roles in their kdapp "my.role" and "my-role" then they probably deserve whatever confusion results.

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/go-logr/logr v0.1.0
 	github.com/google/uuid v1.1.1
 	github.com/operator-framework/operator-sdk v0.15.2
+	github.com/prometheus/common v0.7.0
```
joel-bluedata, line 9: Hm I wonder what pulled this in.

pkg/executor/service.go
joel-bluedata: Hmm there is a length different too. A kdcluster name can be 253 chars while a service name only 63.

We could just pass this through the name munge function. It would be a bit conservative in what it truncates for length, since there is no role ID involved, but that's probably fine.

pkg/executor/util.go
koshiksamota10: Is pod name, statefulset max, pvc name max sizes documented somewhere? I just considered 63 as max for everything and came about with the number 52 with some experimentation. I will document this.

pkg/validator/cluster.go
koshiksamota10: Isnt this in the validation path itself?

pkg/executor/util.go
koshiksamota10: This is where I was trying to get a conversation started because the setting the value to 52 with the assumption that pod name would contain cr name + role name + UID + index, I could never see it in the kubectl get pods output. Bizarre! I know. I tried experimenting a lot and landed up at number 46. I don't have a logical explanation as to why that is the case.

pkg/validator/cluster.go
joel-bluedata: This is our extra webhook-based validation here. The schema validation (handled by the K8s API server) on the other hand should already take care of enforcing the allowed values for the property, since it's declared like this:
```
            namingScheme:
              type: string
              pattern: '^UID$|^CrNameRole$'
```

pkg/executor/util.go
joel-bluedata: You couldn't actually see any pods, or you saw pods that had a different name?

I can try building/deploying this branch myself to see what is going on.

koshiksamota10: I couldn't actually see pods.

joel-bluedata: So if a statefulset doesn't generate pods, there's something amiss in its pod template spec. For example https://github.com/bluek8s/kubedirector/issues/387 would manifest in this way. IIRC you can see what's going on by doing a kubectl describe on the statefulset (to see its events).

pkg/executor/util.go
```diff
@@ -142,3 +147,28 @@ func createPortNameForService(
 	}
 	return strings.ToLower(portInfo.URLScheme) + "-" + portInfo.ID
 }
+
+// MungObjectName is a utility function that truncates the object names
+// to be below nameLengthLimit threshold set for the CrNameRole naming scheme.
```
joel-bluedata, line 152: also replaces dot or underscore with dash

pkg/validator/app.go
```diff
@@ -55,6 +55,16 @@ func (obj appPatchValue) MarshalJSON() ([]byte, error) {
 	return json.Marshal(obj.stringSliceValue)
 }
 
+func validateAppName(
```
joel-bluedata, line 58: IMO we should just add maxLength to the metadata.name in the kdapp CR, rather than dynamically checking in the webhook.

pkg/validator/types.go
```diff
@@ -53,6 +53,7 @@ const (
 	invalidSelectedRoleID = "Invalid element(%s) in selectedRoles array in config section. Valid roles: \"%s\""
 	invalidServiceID      = "Invalid service_id(%s) in roleServices array in config section. Valid services: \"%s\""
 
+	kdappNameLength       = "KDApp name length violates the max name length of 63 characters."
```
joel-bluedata, line 56: Don't need this if we just enforce in the schema in the CRD.

pkg/executor/util.go
```diff
@@ -24,6 +24,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
+// We limit the statefulset, pod, service and pvc name to be lesser than 46
+// characters to accomodate for the controller revision hash and the unique
+// identifier that kubernetes adds while creating the statefulset pods.
```
joel-bluedata, line 29: I had some comments/questions earlier about how we should handle the fact that this number depends on the number of digits in maxKDMembers ... what were your thoughts on that?

joel-bluedata: Also I still think it would be good to have more comments here about how this is calculated. I.e. I've already forgotten enough about the controller revision hash thing to understand why this is 46.

To recap/re-figure-this-out:

I know that one of our limitations is that service names have to be 63 max, and we name our services after the statefulset members which have an index # appended. The length of the GenerateName UID is 5 chars, the current max number of digits in maxKDMembers is 4, plus a hyphen between them, so our GenerateName prefix can be at most 53 to deal with that limitation.

(The statefulset.kubernetes.io/pod-name label value would also probably give us the same sort of limit.)

We also have the controller revision hash label value as a limitation. A label value can be 63 max. Similar calculation as above except instead of an index number there will be a 10 character hash appended. That would seem to indicate that our max size would be 63 - (5 + 1 + 10) = 47.

Since your loop below does "less than" and not "less than or equal to", you're currently using a limit of 45.

Obviously a smaller limit is safe too but the discrepancy does make me think that one of us is overlooking something about what the limit(s) really are. We need to be extra sure we understand what is driving the limits and then add comments here to explain.

pkg/validator/app.go
joel-bluedata: I.e. enhance the current "metadata" part of the kdapp validation to be:
```yaml
        metadata:
          type: object
          properties:
            name:
              type: string
              maxLength: 63
```

pkg/executor/util.go
koshiksamota10: > I had some comments/questions earlier about how we should handle the fact that this number depends on the number of digits in maxKDMembers ... what were your thoughts on that?

Do we have an open issue that talks about changing the member size to be dynamic? If yes, I can add a comment there to address the naming scheme based on the approach that we take there. If not, I can open an issue and add a comment there. As for now, since we are already checking on the current member size, we should be fine? I'll add a comment explaining the max nameLengthLimit calculation.

Let me know if this sounds okay to you.

koshiksamota10: > Also I still think it would be good to have more comments here about how this is calculated. I.e. I've already forgotten enough about the controller revision hash thing to understand why this is 46.
> 
> To recap/re-figure-this-out:
> 
> I know that one of our limitations is that service names have to be 63 max, and we name our services after the statefulset members which have an index # appended. The length of the GenerateName UID is 5 chars, the current max number of digits in maxKDMembers is 4, plus a hyphen between them, so our GenerateName prefix can be at most 53 to deal with that limitation.
> 
> (The statefulset.kubernetes.io/pod-name label value would also probably give us the same sort of limit.)
> 
> We also have the controller revision hash label value as a limitation. A label value can be 63 max. Similar calculation as above except instead of an index number there will be a 10 character hash appended. That would seem to indicate that our max size would be 63 - (5 + 1 + 10) = 47.
> 
> Since your loop below does "less than" and not "less than or equal to", you're currently using a limit of 45.
> 
> Obviously a smaller limit is safe too but the discrepancy does make me think that one of us is overlooking something about what the limit(s) really are. We need to be extra sure we understand what is driving the limits and then add comments here to explain.

I have tried to explain the calculation. Let me know if it makes sense to you.

pkg/executor/util.go
```diff
@@ -24,6 +24,21 @@ import (
...
+// service names, the 5 digit UID and the 4 digit maxKDMember size (1000)
+// should be accounted for.
+// Also, as part of stateful pod creating a 10 digit hash value is added
+// to the controller revision hash label which needs to be accounted for
+// while calculating the prefix size.
+// Naming scheme for the service is as follows: prefix + UID + member index
+// Naming scheme for the label is as follows: prefix + UID + hash value
+// Since, the max member size currently is restricted to be 4 characters, take
+// the max of hash value digits and member size digits which is 10.
+// Prefix calculation is done as following = 63 - 10 - 5 - 2 ('-' characters) = 46.
```
joel-bluedata, line 39: Good explanation! except only one hyphen character will be added, not 2. The first hyphen is part of the prefix whose length we are enforcing.

Also, this calculation determines the maximum valid size of the prefix, i.e. a value that the prefix can be <= to. However the code that references this value enforces that the prefix is < the value rather than <=.

pkg/executor/util.go
joel-bluedata: > Do we have an open issue that talks about changing the member size to be dynamic? If yes, I can add a comment there to address the naming scheme based on the approach that we take there. If not, I can open an issue and add a comment there.

We don't have such an issue; go ahead and create it.

PR #419, opened 2020-09-30
Fix #395: Add guestconfig hooks to connections

### LINKED ISSUE #395: Add guestconfig hooks to "connections"
We introduced the "connections" feature in 0.5 release. Currently when the connected object changes (secret, configmap or cluster) the configmeta inside the pod is updated by we don't have guestconfig hooks so that the apps could react accordingly. 

Add "--reconnect" hook in guestconfig for the same.

saurabh-jogalekar:
### Summary
This feature adds a guestconfig hook when connections change for a kdcluster. Now, when the attached `connections` are changed, the guestconfig/startscript is invoked with `--reconnect` parameter. The guestconfig script can then be configured to use this flag and change the `kdcluster`. Fixes #395 

#### How is this achieved? (high-level)

We count the number of times a `connection` object is changed for every cluster (called `HashChangeIncrementor`), and we extend this concept to every member of that cluster. We count the number of times we call the guestconfig script with `--reconnect` flag for every member (called `LastConnectionVersion`). Whenever `HashChangeIncrementor` > `LastConnectionVersion`, we call guestconfig script and increment `LastConnectionVersion` for the member.   

(Update: We no longer rely on ConnectionsIncrementor since it does not change when the connected object is deleted.)

#### Notes

<s>1. logs for earlier guestconfig invocations are replaced by the latest invocation logs</s> The logs are appended 

- [ ] We Need to add the newly added field in the wiki https://github.com/bluek8s/kubedirector/wiki/KubeDirectorCluster-Definition​​​​​​​ with any newly added


pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -211,6 +211,13 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 	// could cause a handler exit and we would lose the necessary spec gen
 	// update.
 	if state == clusterMembersChangedUnready || (currentHash != cr.Status.LastConnectionHash) {
+
+		shared.LogInfo(
+			reqLogger,
+			cr,
+			shared.EventReasonCluster,
+			fmt.Sprintf("CONNECTIONS INCREMENTOR VALUE IN IF LOOP IS %s", cr.Annotations[shared.ConnectionsIncrementor]),
```
kmathur2, line 219: remove this debug log

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -240,6 +240,43 @@ func handleReadyMembers(
...
+			fmt.Sprintf("In HandleReadyMembers ConnectionsIncrementor value is: %d", connectionsVersion),
+		)
+
+	} else {
+		connectionsVersion = int64(0)
+
+		shared.LogInfo(
+			reqLogger,
+			cr,
+			shared.EventReasonCluster,
```
kmathur2, line 275: remove debug log

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -240,6 +240,43 @@ func handleReadyMembers(
...
+			)
+			return
+
+		}
+
+		shared.LogInfo(
+			reqLogger,
+			cr,
+			shared.EventReasonCluster,
+			fmt.Sprintf("In HandleReadyMembers ConnectionsIncrementor value is: %d", connectionsVersion),
```
kmathur2, line 266: we should only log when the connection changes else reconciler will flood the logs

pkg/controller/kubedirectorcluster/types.go
```diff
@@ -95,6 +95,11 @@ const (
 	2> /opt/guestconfig/configure.stderr  &`
 	fileInjectionCommand = `mkdir -p %s && cd %s &&
 	curl -L %s -o %s`
+	appPrepConfigReconnectCmd = `rm -f /opt/guestconfig/configure.* &&
+	echo -n %s= > ` + appPrepConfigStatus + ` &&
+	nohup sh -c "` + appPrepStartscript + ` --reconnect;
+	echo -n $? >> ` + appPrepConfigStatus + `" > /opt/guestconfig/configure.stdout
```
kmathur2, line 101: let's not overwrite guestconfig logs, I think >> should do

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -401,6 +493,23 @@ func handleCreatingMembers(
 				m.StateDetail.ConfigErrorDetail = errorDetail
 			}
 
+			connectionVersion := int64(0)
+			if v, ok := cr.Annotations[shared.ConnectionsIncrementor]; ok {
+				connectionVersion, _ = strconv.ParseInt(v, 10, 64)
```
kmathur2, line 498: what is the behavior if the annotation is accidentally deleted or update to non string

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -280,10 +317,65 @@ func handleReadyMembers(
...
+
+			shared.LogInfo(
+				reqLogger,
+				cr,
+				shared.EventReasonCluster,
+				fmt.Sprintf("memberVersion for pod %s is %d", m.Pod, memberVersion),
+			)
+
+			if memberVersion < connectionsVersion {
+				shared.LogInfo(
```
kmathur2, line 331: so if the annotation is deleted then we will never come here. Can we infer connectionsVersion from members ?

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -280,10 +317,65 @@ func handleReadyMembers(
...
+					reqLogger,
+					cr,
+					shared.EventReasonCluster,
+					fmt.Sprintf("--reconnect will be called here for pod : %s", m.Pod),
+				)
+
+				containerID := m.StateDetail.LastConfiguredContainer
+				cmd := fmt.Sprintf(appPrepConfigReconnectCmd, containerID)
+
```
kmathur2, line 340: nit: extra space

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -240,6 +240,43 @@ func handleReadyMembers(
...
+			shared.LogErrorf(
+				reqLogger,
+				connVersionError,
+				cr,
+				shared.EventReasonMember,
+				"Invalid connectionsIncrementor for role{%s}",
+				role.roleStatus.Name,
+			)
+			return
+
```
koshiksamota10, line 259: Get rid of the empty line.

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -280,10 +317,65 @@ func handleReadyMembers(
...
+						reqLogger,
+						cmdErr,
+						cr,
+						shared.EventReasonMember,
+						"failed to run startcsript with --reconnect in member{%s} in role{%s}",
+						m.Pod,
+						role.roleStatus.Name,
+					)
+					return
+				}
```
koshiksamota10, line 368: How are we going to handle the failure case here? If the run script fails, the member version would still not be equal to the current connection version, we would retry?

kmathur2:
We Need to add the newly added field in the wiki 
https://github.com/bluek8s/kubedirector/wiki/KubeDirectorCluster-Definition​​​​​​​ with any newly added

pkg/controller/kubedirectorcluster/members.go
saurabh-jogalekar: The behaviour will be consistent with what happens when an error is encountered while injecting configmeta.json in the pod. An error would be logged and the flow would return without retrying again.

pkg/shared/types.go
```diff
@@ -60,3 +60,8 @@ const (
 const (
 	ConnectionsIncrementor = KdDomainBase + "/connUpdateCounter"
 )
+
```
kmathur2, line 63: nit: Please add a comment here clearly mentioning the difference between the 2 counters and why do we need both of them

pkg/controller/kubedirectorcluster/members.go
saurabh-jogalekar: We now have a fall-back mechanism to assign the minimum of the connection version from the ready members.

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -1286,3 +1341,65 @@ func fqdnsList(
...
+
+		return connectionsVersion
+
+	}
+
+	return getDefaultConnectionVersion(reqLogger, cr, role)
+
+}
+
+// getDefaultConnectionVersion returns a conneciton version by parsing the version of members in "ready" state
```
kmathur2, line 1378: typo conneciton

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -1286,3 +1341,65 @@ func fqdnsList(
...
+// otherwise, we return the smallest LastConnectionVersion from all the members.
+// This is because we are better off running --reconnect when no connection has changed, rather than not running --reconnect when a connection change does happen
+func getDefaultConnectionVersion(
+	reqLogger logr.Logger,
+	cr *kdv1.KubeDirectorCluster,
+	role *roleInfo,
+) int64 {
+
+	ready := role.membersByState[memberReady]
+
```
kmathur2, line 1389: nit: remove empty

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -1286,3 +1341,65 @@ func fqdnsList(
...
+	if len(ready) == 0 {
+		x := int64(0)
+		return x
+	}
+
+	min := int64(math.MaxInt64)
+	for _, memberStatus := range ready {
+		memberVersion := *memberStatus.StateDetail.LastConnectionVersion
+		if min > memberVersion {
```
kmathur2, line 1398: min is math.MaxInt64, this will be always true?

kmathur2: also why max is named min, looks odd

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -1286,3 +1341,65 @@ func fqdnsList(
...
+// we get the connection version from members that are in "Ready" state
+func getConnectionVersion(
+	reqLogger logr.Logger,
+	cr *kdv1.KubeDirectorCluster,
+	role *roleInfo,
+) int64 {
+
+	if connectionsVersionStr, ok := cr.Annotations[shared.HashChangeIncrementor]; ok {
+		connectionsVersion, connVersionError := strconv.ParseInt(connectionsVersionStr, 10, 64)
+
```
kmathur2, line 1356: remove empty

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -1286,3 +1341,65 @@ func fqdnsList(
...
+			shared.LogErrorf(
+				reqLogger,
+				connVersionError,
+				cr,
+				shared.EventReasonMember,
+				"Invalid connectionsIncrementor for role{%s}",
+				role.roleStatus.Name,
+			)
+			return getDefaultConnectionVersion(reqLogger, cr, role)
+
```
kmathur2, line 1367: remove empty

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -1286,3 +1341,65 @@ func fqdnsList(
...
+		}
+
+		return connectionsVersion
+
```
kmathur2, line 1371: remove empty

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -1286,3 +1341,65 @@ func fqdnsList(
...
+	}
+
```
kmathur2, line 1373: remove empty

pkg/controller/kubedirectorcluster/members.go
saurabh-jogalekar: We assume a big value(here max int64 value) as min and then assign min if that value > member version. This will get us a minimum member value.

kmathur2: if min = math.Max then math.Max will always be  > memberVersion right? Am I missing something  ?

kmathur2 (approved):

PR #421, opened 2020-10-07
App Logos [Issue #423]

koshiksamota10:
https://github.com/bluek8s/kubedirector/issues/423
Add a field to the app crd where we can store URL path for logos associated with a specific app.

### DIFF
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
@@ -205,7 +205,7 @@ spec:
                 minLength: 1
             systemdRequired:
               type: boolean
-            defaultLogoURL:
+            logoURL:
               type: string
               minLength: 1
               pattern: '^(file|https|http?)://.+\.(jpeg|png)$'
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
@@ -205,3 +205,7 @@ spec:
                 minLength: 1
             systemdRequired:
               type: boolean
+            defaultLogoURL:
+              type: string
+              minLength: 1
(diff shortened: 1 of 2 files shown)


deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
```diff
@@ -204,4 +204,8 @@ spec:
                 type: string
                 minLength: 1
             systemdRequired:
-              type: boolean
\ No newline at end of file
+              type: boolean
+            defaultLogoURL:
+              type: string
+              minLength: 1
+              pattern: '^(file|https?)://.+\.tgz$'
```
kmathur2, line 211: should we also allow http ?

kmathur2: logo should not end with tgz, png/jpeg or other popular image formats

pkg/apis/kubedirector/v1beta1/kubedirectorapp_types.go
```diff
@@ -34,6 +34,7 @@ type KubeDirectorAppSpec struct {
 	DefaultEventList    *[]string           `json:"defaultEventList,omitempty"`
 	Capabilities        []corev1.Capability `json:"capabilities,omitempty"`
 	SystemdRequired     bool                `json:"systemdRequired,omitempty"`
+	DefaultLogoURL      string              `json:"defaultLogoURL,omitempty"`
```
saurabh-jogalekar, line 37: [nit] wondering if we could simply name this field as `LogoURL`. The other `Default` fields like `DefaultImageRepoTag`, `DefaultSetupPackage` ... are also applicable at member level. Since logo would be a KDApp-specific attribute (like `Version`), could we also keep the logo field name consistent with this scheme?

koshiksamota10: I think it makes sense. Will change it.

kmathur2 (approved):

saurabh-jogalekar (approved):
LGTM

PR #441, opened 2020-12-04
Docker file update for LDAP updates and BDP2820 #581

arpitarathi:
The change will :
1. remove extra R Kernel,
2. use LDAP authenticator and therefore allow users with underscore,
3. direct bind support


kmathur2:
Looks like this is redundant. Closing

PR #443, opened 2021-01-09
Add volumeDevice support for KDApps

### LINKED ISSUE #442: Recurring log message in pod logs
The following line is frequently seen in the pod logs:

`"Initialized Annotations and updated context"`

Consider removing the log entry.

saurabh-jogalekar:
### Summary
- Implementation for issue #444
- Support to mount block devices.
- If such block device(s) is/are mounted, add the mount path(s) in appconfig JSON

- Fix #442  

### What's new?

The Cluster Spec is updated to have `BlockStorage` field which takes `prefixPath`, `numDevices`, and `storageClassName` for mounting block(s) to the role's statefulset. `size` is an optional parameter and a default size of `1Gi` is set if it is not specified. 

`storageClassName` will be used to bind an appropriate PV. The `PV` and `StorageClass` need to be created before launching a KDApp with block devices. 

~~Currently, the code supports one deviceVolume which is commensurate with volume mount support.~~ The code now supports mounting multiple blocks. `numDevices` reads the number of blocks required, and `prefixPath` reads the template to the path where the device is to be mounted. 

After the block device(s) are added, the mount path is registered in appconfig JSON (configmeta.json) 
The path is registered as `block_device_path` sub-field in the `node` parent field.

#### Sample `centos` App with block device and volume mount:

```yaml
apiVersion: "kubedirector.hpe.com/v1beta1"
kind: "KubeDirectorCluster"
metadata: 
  name: "centos7x-instance"
  namespace: "pvc-test-4"
spec: 
  app: "centos7x"
  appCatalog: "local"
  roles: 
    - 
      id: "vanilla_centos"
      members: 1
      resources: 
        requests: 
          memory: "4Gi"
          cpu: "2"
        limits: 
          memory: "4Gi"
          cpu: "2"
      storage: 
        size: "1Gi"
      blockStorage:
        pathPrefix: "/dev/xdb"
        storageClassName: "kd-block-storage"
        numDevices: 2
``` 

Corresponding `node` field in the App's `configmeta.json`:

```json
...
...
"node": {
        "role_id": "vanilla_centos",
        "nodegroup_id": "1",
        "id": "1",
        "hostname": "kdss-tl5zz-0.kdhs-4dnfp.seldon-t.svc.cluster.local",
        "fqdn": "kdss-tl5zz-0.kdhs-4dnfp.seldon-t.svc.cluster.local",
        "domain": "kdhs-4dnfp.seldon-t.svc.cluster.local",
        "distro_id": "hpecp/centos7",
        "depends_on": {}
        "block_device_path": ["/dev/xdb0", "/dev/xdb1"] 
    },
...
...
``` 

Sample `PersistentVolume` which uses `/dev/sdb` path on the specified host as a block when bound to a statefulset can be like the following. Note that if `n` blocks are to be mounted for a member, there should at least `n` PVs on one of the cluster nodes:
```yaml

apiVersion: v1
kind: PersistentVolume
metadata:
  name: cldb1
spec:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 1Gi
  local:
    path: /dev/sdb
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: kubernetes.io/hostname
          operator: In
          values:
          - my-host.net
  persistentVolumeReclaimPolicy: Retain
  storageClassName: kd-block-storage
  volumeMode: Block
```

Sample `StorageClass`:

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: kd-block-storage
provisioner: kubernetes.io/no-provisioner
volumeBindingMode: Immediate
reclaimPolicy: Delete
```

*Note* the `volumeBindingMode` and `reclaimPolicy` values are for illustration. One may or may not need to alter them for the Volumes to bind.

### DIFF
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
@@ -169,6 +169,19 @@ spec:
                       storageClassName:
                         type: string
                         minLength: 1
+                  blockStorage:
+                    type: object
+                    nullable: true
+                    required: [storageClassName, path]
+                    properties:
+                      path:
+                        type: string
+                      storageClassName:
+                        type: string
+                        minLength: 1
+                      size:
+                        type: string
+                        pattern: '^([0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
                   fileInjections:
                     type: array
                     items:
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
@@ -258,6 +258,8 @@ spec:
                           type: string
                         pvc:
                           type: string
+                        blockdevicepath:
+                          type: string
                         authToken:
                           type: string  
                         state:
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
@@ -180,7 +180,7 @@ spec:
                       pathPrefix:
                         type: string
                         minLength: 1
-                        pattern: '^/.*$' #start with '/'. 
+                        pattern: '^/.*$'
                       storageClassName:
                         type: string
                         minLength: 1
@@ -263,7 +263,7 @@ spec:
                           type: string
                         pvc:
                           type: string
-                        blockDevicePath:
+                        blockDevicePaths:
                           type: array
                         authToken:
                           type: string  
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
@@ -180,6 +180,7 @@ spec:
                       pathPrefix:
                         type: string
                         minLength: 1
+                        pattern: '^/.*$' #start with '/'. 
                       storageClassName:
                         type: string
                         minLength: 1
@@ -262,7 +263,7 @@ spec:
                           type: string
                         pvc:
                           type: string
-                        blockdevicepath:
+                        blockDevicePath:
                           type: array
                         authToken:
                           type: string  
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
@@ -172,10 +172,14 @@ spec:
                   blockStorage:
                     type: object
                     nullable: true
-                    required: [storageClassName, path]
+                    required: [storageClassName, pathPrefix, numDevices]
                     properties:
-                      path:
+                      numDevices:
+                        type: integer
+                        minimum: 1
+                      pathPrefix:
                         type: string
+                        minLength: 1
                       storageClassName:
                         type: string
                         minLength: 1
@@ -259,7 +263,7 @@ spec:
                         pvc:
                           type: string
                         blockdevicepath:
-                          type: string
+                          type: array
                         authToken:
                           type: string  
                         state:
--- a/pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
+++ b/pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
@@ -135,6 +135,7 @@ type Role struct {
 	EnvVars        []corev1.EnvVar             `json:"env,omitempty"`
 	FileInjections []FileInjections            `json:"fileInjections,omitempty"`
 	Secret         *KDSecret                   `json:"secret,omitempty"`
+	BlockStorage   *BlockStorage               `json:"blockStorage,omitempty"`
 }
 
 // StateRollup surfaces whether any per-member statuses have problems that
@@ -153,6 +154,14 @@ type StateRollup struct {
 type ClusterStorage struct {
 	Size         string  `json:"size"`
 	StorageClass *string `json:"storageClassName,omitempty"`
+}
+
+// BlockStorage defines the block storage type and quantity, if any, to be used
+// for mounting a block volume in a role.
+type BlockStorage struct {
+	StorageClass *string `json:"storageClassName,omitempty"`
+	Path         *string `json:"path,omitempty"`
+	Size         *string `json:"size,omitempty"`
 }
 
 // RoleStatus describes the component objects of a virtual cluster role.
--- a/pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
+++ b/pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
@@ -174,14 +174,14 @@ type RoleStatus struct {
 
 // MemberStatus describes the component objects of a virtual cluster member.
 type MemberStatus struct {
-	Pod             string            `json:"pod"`
-	Service         string            `json:"service"`
-	AuthToken       string            `json:"authToken,omitempty"`
-	PVC             string            `json:"pvc,omitempty"`
-	State           string            `json:"state"`
-	StateDetail     MemberStateDetail `json:"stateDetail,omitempty"`
-	NodeID          int64             `json:"nodeID"`
-	BlockDevicePath []string          `json:"blockDevicePath,omitempty"`
+	Pod              string            `json:"pod"`
+	Service          string            `json:"service"`
+	AuthToken        string            `json:"authToken,omitempty"`
+	PVC              string            `json:"pvc,omitempty"`
+	State            string            `json:"state"`
+	StateDetail      MemberStateDetail `json:"stateDetail,omitempty"`
+	NodeID           int64             `json:"nodeID"`
+	BlockDevicePaths []string          `json:"blockDevicePaths,omitempty"`
 }
 
 // MemberStateDetail digs into detail about the management of configmeta and
--- a/pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
+++ b/pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
@@ -156,7 +156,7 @@ type ClusterStorage struct {
 	StorageClass *string `json:"storageClassName,omitempty"`
 }
 
-// BlockStorage defines the block storage type and quantity, if any, to be used
+// BlockStorage defines the block storage type, path, and optionally size, if any, to be used
 // for mounting a block volume in a role.
 type BlockStorage struct {
 	StorageClass *string `json:"storageClassName,omitempty"`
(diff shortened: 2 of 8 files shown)


saurabh-jogalekar:
@kmathur2 @swamibluedata could you take a look?

kmathur2:
We should also add the info in config metadata for apps to be able to consume it

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
```diff
@@ -155,6 +156,14 @@ type ClusterStorage struct {
 	StorageClass *string `json:"storageClassName,omitempty"`
 }
 
+// BlockStorage defines the block storage type, path, and optionally size, if any, to be used
+// for mounting a block volume in a role.
+type BlockStorage struct {
+	StorageClass *string `json:"storageClassName,omitempty"`
+	Path         *string `json:"path,omitempty"`
```
kmathur2, line 163: nit: let's keep this consistent with the property name in the status which is blockdevicepath

pkg/executor/types.go
```diff
@@ -70,6 +70,11 @@ const (
 	// nvidiaGpuVisWorkaroundEnvVarValue is the value to be set for the environment variable
 	// named nvidiaGpuVisWorkaroundEnvVarName, in the above work-around
 	nvidiaGpuVisWorkaroundEnvVarValue = "VOID"
+	// defaultBlockDeviceSize is the size for a block volume if it is not specified in the spec
+	defaultBlockDeviceSize = "1Gi"
```
kmathur2, line 74: is this size just for the API or is actually useful?

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
```diff
@@ -155,6 +156,14 @@ type ClusterStorage struct {
...
+	Size         *string `json:"size,omitempty"`
```
kmathur2, line 164: do we need size if we are always defaulting?

pkg/executor/statefulset.go
```diff
@@ -450,9 +461,46 @@ func getVolumeClaimTemplate(
...
-	return
+
+	if role.BlockStorage != nil {
+
+		block := v1.PersistentVolumeBlock
+
+		blockVolSize, _ := resource.ParseQuantity(defaultBlockDeviceSize)
+
+		if role.BlockStorage.Size != nil {
+			blockVolSize, _ = resource.ParseQuantity(*role.BlockStorage.Size)
```
kmathur2, line 475: is it useful? can we not always default?

deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
```diff
@@ -169,6 +169,19 @@ spec:
                       storageClassName:
                         type: string
                         minLength: 1
+                  blockStorage:
+                    type: object
+                    nullable: true
+                    required: [storageClassName, path]
+                    properties:
+                      path:
+                        type: string
```
kmathur2, line 178: need regex/validation for path

kmathur2:
Overall looks good, what happens if storageClass/PV is not present at cluster launch time?  It should be ok and pods should stay in the pending state, right?

deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
```diff
@@ -169,6 +169,19 @@ spec:
...
+                      storageClassName:
+                        type: string
+                        minLength: 1
+                      size:
+                        type: string
+                        pattern: '^([0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
```
swamibluedata, line 184: we also talked about numDevices here?

deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
```diff
@@ -245,6 +258,8 @@ spec:
                           type: string
                         pvc:
                           type: string
+                        blockdevicepath:
```
swamibluedata, line 261: nit: blockDevicePath

deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
swamibluedata: Hmm this should be changed when we add numDevices. What should reflect is just the prefix path, and we should incremental numbers. So use can say something like "/dev/kdb" and inside the container it should be /dev/kdb0, /dev/kdb1 etc

swamibluedata:
Looks good, you do need to change members.go to delete pvc when statefulset is deleted. Can you double check

deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
saurabh-jogalekar: `pathPrefix` now requires a minimum of one character. I think it is on the user to specify a valid path. Thoughts?

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
saurabh-jogalekar: @swamibluedata : Irrespective of what size we specify, the block size always remains the same. Any use case when this might be useful?

saurabh-jogalekar:
@kmathur2 , @swamibluedata : Updated the code to support multiple device mounts. The only open issue that I see is whether or not size should be an attribute in the Spec.

saurabh-jogalekar:
> Looks good, you do need to change members.go to delete pvc when statefulset is deleted. Can you double check

All PVCs(mounts and devices) get deleted without any additional code.

deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
kmathur2: what will be the status of cluster if invalid path is provided?

saurabh-jogalekar: Pod goes in Error state:

```
[root@mip-bd-vm226 kd-pvc]# kubectl get pod
NAME           READY   STATUS                       RESTARTS   AGE
kdss-2rglb-0   0/1     CreateContainerConfigError   0          52s
```


Describe yields:
```
  Warning  Failed                 6s (x7 over 66s)  kubelet, mip-bd-vm226.mip.storage.hpecorp.net  Error: error DevicePath `invalid0` must be an absolute path
```

swamibluedata: That makes sense. First character has to be "/"

We can mandate this to be /dev/kdb or something like that also

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
swamibluedata: block size inside the pod should reflect the size of the underlying block device?

deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
saurabh-jogalekar: Okay. I shall add a regex check to ensure that the first character is `/`
Sounds good?

kmathur2: I think that's fine for now.

pkg/executor/statefulset.go
```diff
@@ -450,9 +474,55 @@ func getVolumeClaimTemplate(
...
+		}
+
+		numDevices := *role.BlockStorage.NumDevices
+		var i int32
```
kmathur2, line 492: nit: don't need to explicitly declare i, can always use :=

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
saurabh-jogalekar: From what I found, a block volume occupies the entire partition/block of storage. Storage size in this context would be helpful when we check if the Claim Template Size is less than or equal to the storage capacity in PV.
eg: if PV is defined with storage capacity of 1Gi and the claim template requests 10Gi, the PV wont bind.

saurabh-jogalekar: I think we can keep it optional and modifiable by the user if, in case, the default value does not work.

kmathur2:
Let's update the docs also here https://github.com/bluek8s/kubedirector/wiki/KubeDirectorCluster-Definition

kmathur2 (approved):

pkg/catalog/types.go
```diff
@@ -57,14 +57,16 @@ type cluster struct {
...
+	RoleID          string     `json:"role_id"`
+	NodegroupID     string     `json:"nodegroup_id"`
+	ID              string     `json:"id"`
+	Hostname        string     `json:"hostname"`
+	FQDN            string     `json:"fqdn"`
+	Domain          string     `json:"domain"`
+	DistroID        string     `json:"distro_id"`
+	DependsOn       refkeysMap `json:"depends_on"`
+	BlockDevicePath []string   `json:"block_device_path,omitempty"`
+	// we do not want the field if no block device is attached. Hence omitempty
```
swamibluedata, line 69: do we need this comment?

deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
```diff
@@ -169,6 +169,24 @@ spec:
...
+                    nullable: true
+                    required: [storageClassName, pathPrefix, numDevices]
+                    properties:
+                      numDevices:
+                        type: integer
+                        minimum: 1
+                      pathPrefix:
+                        type: string
+                        minLength: 1
+                        pattern: '^/.*$' #start with '/'.
```
swamibluedata, line 183: remove comment

deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
```diff
@@ -245,6 +263,8 @@ spec:
                           type: string
                         pvc:
                           type: string
+                        blockDevicePath:
```
swamibluedata, line 266: more nit: may be blockDevices or blockDevicePaths

pkg/catalog/configmeta.go
```diff
@@ -467,15 +467,17 @@ func ConfigmetaGenerator(
...
-				DependsOn:   make(refkeysMap), // currently, always empty
+				RoleID:          roleName,
+				NodegroupID:     "1",
+				ID:              strconv.FormatInt(member.NodeID, 10),
+				Hostname:        memberName + "." + domain,
+				FQDN:            memberName + "." + domain,
+				Domain:          domain,
+				DistroID:        appCR.Spec.DistroID,
+				DependsOn:       make(refkeysMap),       // currently, always empty
+				BlockDevicePath: member.BlockDevicePath, // present only if block device is to be mounted
```
swamibluedata, line 480: here too

swamibluedata:
Looks great. just minor comments

PR #456, opened 2021-03-22
Bluek8s ConfluentKafka app name changed to ConfluentPlatform

arpitarathi:
EZEPIC-22 - Need to rename "Confluent Kafka" in the app store - equivalent changes in KD App for Confluent Platform:
Changes done on App json, Cluster yaml definitions and documentation to reflect Confluent Platform instead of Confluent Kafka in app description

### DIFF
--- a/deploy/example_catalog/cr-app-kafka55.json
+++ b/deploy/example_catalog/cr-app-kafka55.json
@@ -2,7 +2,7 @@
     "apiVersion": "kubedirector.hpe.com/v1beta1",
     "kind": "KubeDirectorApp",
     "metadata": {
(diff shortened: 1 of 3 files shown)


joel-bluedata (approved):
OK by me if it's OK by Kartik.

(In this case, we don't need to update the top-level "version" since we are changing the distro ID.)

kmathur2 (approved):

kmathur2:
> OK by me if it's OK by Kartik.
> 
> (In this case, we don't need to update the top-level "version" since we are changing the distro ID.)

Sure !

Issue #48, opened 2018-10-03
placement constraints


joel-bluedata:
Think about how we could/should expose affinity and nodeSelector configuration for the deployed pods.

joel-bluedata:
Current thinking is that node selector, affinity, and labelling for pods are all things that could need to be enforced in the kdapp and/or requested in the kdcluster.

We currently have support for the "requested in the kdcluster" part of that story for labels. We should probably extend that to node selector and affinity, and also (for all those things) look into expressing them in the kdapp as well.

Vellanci:
As discussed with @kmathur2 the implementation for now will be adding affinity field as part of RoleSpec specified in kdclusters and passing it to pods. Also a good idea is to add affinity to list of pods in kdclusters' status. The affinity is more advanced and therefore top priority for this feature. Node Selector and Node Name fields can be added later as well.

PR #458, opened 2021-04-08
#48 Allow specifying Affinity for roles

Vellanci:
As requested in issue #48:

- add Affinity field to KDCluster's Role spec
- use it to define affinity for corresponding StatefulSet
- update CRDs

### DIFF
--- a/pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
+++ b/pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
@@ -131,6 +131,7 @@ type Role struct {
 	ServiceLabels  map[string]string           `json:"serviceLabels,omitempty"`
 	Members        *int32                      `json:"members,omitempty"`
 	Resources      corev1.ResourceRequirements `json:"resources"`
+	Affinity       *corev1.Affinity            `json:"affinity,omitempty"`
 	Storage        *ClusterStorage             `json:"storage,omitempty"`
 	EnvVars        []corev1.EnvVar             `json:"env,omitempty"`
 	FileInjections []FileInjections            `json:"fileInjections,omitempty"`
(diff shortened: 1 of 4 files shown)


pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
```diff
@@ -182,6 +183,7 @@ type MemberStatus struct {
 	StateDetail      MemberStateDetail `json:"stateDetail,omitempty"`
 	NodeID           int64             `json:"nodeID"`
 	BlockDevicePaths []string          `json:"blockDevicePaths,omitempty"`
+	Affinity         *corev1.Affinity  `json:"affinity,omitempty"`
```
joel-bluedata, line 186: What do we gain by copying this into the member status?

kmathur2:
@Vellanci As a general practice, please add some description about the PR and changes.

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
kmathur2: Good point, we should record the affinity for a given member wrt affinity, right ? Just for book keeping

joel-bluedata: It will always be the same as the role config (since the role config can't be modified while members exist), so I don't think we need to duplicate it here.

It might provide some convenience to have it in the member status, but I think we have to be cautious about expanding the size of the member status since there can be a LOT of these and we have an overall size limit for the CR.

Vellanci: Makes sense to me. Removed Affinity from status

kmathur2 (approved):

neogeographica:
Yep. Trusting that the schema is correct. :-) I'll update the wiki.

neogeographica (approved):

joel-bluedata:
Sorry, was logged in as my other GitHub account. Yeah it looks ok now. 👍 Trusting that the schema is correct!

I'll update the wiki.

joel-bluedata (approved):

Issue #470, opened 2021-05-06
manual deletion of kdcluster-backing statefulset causes sset spam


joel-bluedata:
Regression in some of the reconcile code. I think the issue is in handleRoleReCreate in roles.go; we need to make sure that if we are creating another statefulset when the roleStatus already exists, we must update roleStatus.StatefulSet to point to the newly created set. This might be as simple as setting role.roleStatus to nil before calling handleRoleCreate.

PR #471, opened 2021-05-06
fix handling of out-of-band-deleted statefulsets

joel-bluedata:
closes issue #470

Two changes here:

- In handleRoleCreate, if the role status already exists, we need to update the statefulset name in the role status. This properly handles the case where a statefulset is "deleted out from under us" and we need to re-create it. If we don't record the name of the new replacement statefulset we get caught in Groundhog Day always trying to handle the missing original set.

- In handleRoleReCreate, don't set anyMembersChanged to true if we didn't actually change any member statuses. This prevents the reconciler from being rapidly called over and over again between the initial detection of the missing statefulset and the creation of the replacement one... basically a cosmetic issue but the KD logspam should be avoided.


kmathur2:
Looks good. Thanks for taking care of this Joel

kmathur2 (approved):

Issue #466, opened 2021-04-30
failure if fractional CPU resource requested for kdcluster


joel-bluedata:
Observed on K8s 1.20.2.

If I try to create a kdcluster with a "0.5" CPU resource, it doesn't reach a final configured state. The process is blocked with these errors in the KD log:

{"level":"error","ts":1619800365.0318282,"logger":"controller_kubedirectorcluster","msg":"trying status update again in 16s; failed","Request.Namespace":"import-test","Request.Name":"centos8","error":"admission webhook \"validate-cr.kubedirector.hpe.com\" denied the request: \nChange to spec not allowed before previous spec change has been processed.", ... }

If I enable more detailed kdcluster logging for the API server, I see that the problematic request includes a spec which is changing the CPU resource from "0.5" to "500m". Obviously this is quantitatively equivalent, but it triggers the "is the spec being changed" check in the validator which is just doing an object comparison.

If I just use "500m" in the initial creation, everything works fine.

A few questions at this point:

- Is this behavior new to 1.20, or have we just not seen the issue before because people tend to request whole-number CPU resource values?

- The KD code is writing directly to the CR's status subresource, so why is the API server receiving a PUT to the CR's spec? The rejection of this PUT is being seen by the status write though, so the spec update is happening somewhere along the code path for that. Maybe in the controller-runtime client code?

- Where is this resource value getting changed from "0.5" to "500m"? We're not doing it explicitly in the KD code. Again maybe something further down that status update callstack.

We could work around this issue by adding logic to KD, probably in the block of code currently marked by "Shortcut out of here if the spec is not being changed" in admitClusterCR. We could specifically dig into the resource properties, see if they are quantitatively equivalent, and if so then copy the new properties into the old object before doing the object compare.

But it would be nice to understand a bit more about what is happening.

PR #473, opened 2021-05-12
resource quantity format changes shouldn't trigger "spec has changed" check

joel-bluedata:
There are situations where a client is not allowed to change a role spec: if the previous change is still being processed, or if the role has nonzero member count.

As per issue #466 though, there are some cases where a resource quantity is getting its format changed, e.g. from "0.5" to "500m", without its quantity actually being changed. Sometimes this is even happening as part of a status write from KD itself. We're not explicitly asking for this format change -- something along the way in client-go or K8s is doing it -- but whatever the reason, we need to allow it to happen.

So this change to how we handle checking for CR changes is motivated by issue #466, but really it applies everywhere we do such a check.

Along with general regression-test creations of existing kdapp types while monitoring the KD logs, I did these specific tests:

- creating kdcluster with 0.5 CPU (works now)
- changing kdcluster role spec to something "semantically equivalent", for role with existing members (works now)
- changing kdcluster role spec to something non-equivalent, for role with existing members (blocked, as it should be)
- changing the members count for a scaleout kdcluster role (works, as it should)
- manually editing a kdcluster status stanza (blocked, as it should be)
- manually editing a kdconfig status stanza (blocked, as it should be)

My current test K8s/Ezmeral system is pretty old, so I'm going to do a clean install to get up-to-date then test one more time.

### DIFF
--- a/pkg/validator/cluster.go
+++ b/pkg/validator/cluster.go
@@ -19,7 +19,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"reflect"
 	"strconv"
 	"strings"
 	"time"
@@ -30,6 +29,7 @@ import (
 	"github.com/bluek8s/kubedirector/pkg/observer"
 	"github.com/bluek8s/kubedirector/pkg/shared"
 	"k8s.io/api/admission/v1beta1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	appsvalidation "k8s.io/apimachinery/pkg/apis/meta/v1/validation"
@@ -390,7 +390,7 @@ func validateRoleChanges(
 		// the new spec if anything other than the members count is different.
 		compareRole := *role
 		compareRole.Members = prevRole.Members
-		if !reflect.DeepEqual(&compareRole, prevRole) {
+		if !equality.Semantic.DeepEqual(&compareRole, prevRole) {
 			roleModifiedMsg := fmt.Sprintf(
 				modifiedRole,
 				role.Name,
@@ -890,7 +890,7 @@ func admitClusterCR(
 				return &admitResponse
 			}
 		} else {
-			if !reflect.DeepEqual(clusterCR.Status, prevClusterCR.Status) {
+			if !equality.Semantic.DeepEqual(clusterCR.Status, prevClusterCR.Status) {
 				admitResponse.Result = statusViolation
 				return &admitResponse
 			}
@@ -899,7 +899,7 @@ func admitClusterCR(
 		// it's OK to write the status again as long as nothing is changing.
 		// (For example we'll see this when a PATCH happens.)
 		if expectedStatusGen.Validated {
-			if !reflect.DeepEqual(clusterCR.Status, prevClusterCR.Status) {
+			if !equality.Semantic.DeepEqual(clusterCR.Status, prevClusterCR.Status) {
 				admitResponse.Result = statusViolation
 				return &admitResponse
 			}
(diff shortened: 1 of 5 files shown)


pkg/validator/cluster.go
```diff
@@ -832,6 +833,64 @@ func validateNamingScheme(
...
+		prevLimits := prev.Resources.Limits
+		newLimits := new.Resources.Limits
+		if !normalizeResourceQuantities(prevLimits, newLimits) {
+			return
+		}
+		prevRequests := prev.Resources.Requests
+		newRequests := new.Resources.Requests
+		if !normalizeResourceQuantities(prevRequests, newRequests) {
+			return
+		}
```
Vellanci, line 890: Wouldn't it be easier to make equality function for ResourceList and then simply assigning prev.Resources.XXX to new.Resources.XXX?

joel-bluedata: Maybe! Hmm. I think the equality function would have to look pretty much like the current code, just without the assignment that is currently happening when the equality is detected. So I'm not sure it would result in less code.

pkg/validator/cluster.go
```diff
@@ -914,6 +973,10 @@ func admitClusterCR(
 	// metadata generation number here because that is incremented after this
 	// validator sees the request.
 	if ar.Request.Operation == v1beta1.Update {
+		normalizeRolesResourceQuantities(
+			prevClusterCR.Spec.Roles,
+			clusterCR.Spec.Roles,
+		)
 		if reflect.DeepEqual(clusterCR.Spec, prevClusterCR.Spec) {
```
Vellanci, line 980: How about using equality.Semantic.DeepEqual instead? It already fixes the issue with Quantities (see https://pkg.go.dev/k8s.io/apimachinery/pkg/api/equality)

pkg/validator/cluster.go
```diff
@@ -832,6 +833,64 @@ func validateNamingScheme(
 	return patches
 }
 
+// normalizeResourceQuantities is called by normalizeRolesResourceQuantities
+// to handle each role resource list. If the quantity of a resource is
+// equivalent between old and new lists, then copy the new value into the old
+// list. If any resources have been removed or added, this function will
+// bail out early returning false; otherwise it returns true once done.
+func normalizeResourceQuantities(
+	prevList corev1.ResourceList,
```
kmathur2, line 842: don't know why they call this ResourceList it should be ResourceMap :)

kmathur2:
Looks good to me 👍

pkg/validator/cluster.go
kmathur2: This is CR spec comparison and not just quantity

kmathur2 (approved):

pkg/validator/cluster.go
Vellanci: > This is CR spec comparison and not just quantity

Yes, I understand it. And semantic equality was designed for these kind of situations. It works same as reflect.DeepEquals but compares objects like Quantity, Time etc by their equality functions

pkg/validator/cluster.go
joel-bluedata: yah

pkg/validator/cluster.go
joel-bluedata: Neat! I'll have a look.

kmathur2: If this works then it will take care of _storage_ field in role spec also. @joel-bluedata we should do the same for storage size also, right?

joel-bluedata: Probably yes. I'm going to see if all of our uses of reflect.DeepEqual should just be equality.Semantic.DeepEqual instead.

joel-bluedata:
OK I changed all of our uses of reflect.DeepEqual. Updated the PR description with details.

kmathur2:
> OK I changed all of our uses of reflect.DeepEqual. Updated the PR description with details.

Neat !! Thanks, @Vellanci for digging the semantic check.

Vellanci (approved):
LGTM

PR #474, opened 2021-05-12
Fix #467 - nil deref in HandleReadyMembers

### LINKED ISSUE #467: possibility of nil deref in handleReadyMembers
In the member state detail, LastConnectionVersion (a new thing introduced in recent KD versions) is declared as a pointer-to-int, so if it is not present in the doc then it will be nil in the code.

This field is entirely handled within members.go.

It's first set in handleCreatingMembers.

Later, it's dereferenced in handleReadyMembers -- either directly, or by calling getConnectionVersion (which calls getDefaultConnectionVersion). This dereference is not protected against a possibility of a nil value.

Since creating status comes before ready status, that's not a problem in the normal way of things. However I'm concerned about what will happen if KD is updated to the latest operator version while previously-created kdclusters exist. kdclusters created under old versions of KD will not have this field in their member state yet, and so it looks like we would get a nil deref whenever we first try to do handleReadyMembers on such a kdcluster.

saurabh-jogalekar:
### Summary:

This issue is encountered in the following scenario:

When a KD Cluster does not contain `lastConnectionVersion` member status (will be typically found in v0.5.1 and below) and the connections are updated. This leads to `handleReadyMembers()` checking for the `memberVersion` (which is nil) and as a result, KD goes in panic state. 

### Proposed Change:

`handleReadyMembers()` gets called only when we are operating on ready members. Currently, it is called to inject configmeta initially and thereafter, every time the connections are modified. We also use this function to call startscript with --reconnect flag only if the connections are modified. 

In the scenario when the deref is encountered, the only possible way to enter `handleReadyMembers()` is when connections are modified. (since the KDCluster is already up) 

Now, we want to call the startscript with --reconnect and initialize the `memberVersion`. Setting it to `connectionVersion - 1` would help both the causes since the flow will call startscript and after that, increment `memberVersion` by 1 and make it equal to `connectionVersion`.

### DIFF
--- a/pkg/controller/kubedirectorcluster/members.go
+++ b/pkg/controller/kubedirectorcluster/members.go
@@ -297,7 +297,7 @@ func handleReadyMembers(
 				shared.LogInfof(
 					reqLogger,
 					cr,
-					"",
+					shared.EventReasonNoEvent,
 					"LastConnectionVersion for {%s} is nil. Resetting.",
 					m.Pod,
 				)
--- a/pkg/controller/kubedirectorcluster/members.go
+++ b/pkg/controller/kubedirectorcluster/members.go
@@ -297,7 +297,7 @@ func handleReadyMembers(
 				shared.LogInfof(
 					reqLogger,
 					cr,
-					shared.EventReasonCluster,
+					"",
 					"Member Version for {%s} is nil. Resetting.",
 					m.Pod,
 				)
--- a/pkg/controller/kubedirectorcluster/members.go
+++ b/pkg/controller/kubedirectorcluster/members.go
@@ -298,7 +298,7 @@ func handleReadyMembers(
 					reqLogger,
 					cr,
 					"",
-					"Member Version for {%s} is nil. Resetting.",
+					"LastConnectionVersion for {%s} is nil. Resetting.",
 					m.Pod,
 				)
 
--- a/pkg/controller/kubedirectorcluster/members.go
+++ b/pkg/controller/kubedirectorcluster/members.go
@@ -284,20 +284,26 @@ func handleReadyMembers(
 				return
 			}
 
+			// memberVersion is the connectionVersion of the member.
+			// If it is nil, we set it to connectionsVersion - 1. This is to call startscript with --reconnect and then equalize connectionsVersion and memberVersion
+			// If it is not nil, we set it to the member's connectionVersion
 			var memberVersion int64
 
 			if m.StateDetail.LastConnectionVersion != nil {
 				memberVersion = *m.StateDetail.LastConnectionVersion
 
 			} else {
-				memberVersion = connectionsVersion - 1
 
-				shared.LogInfo(
+				shared.LogInfof(
 					reqLogger,
 					cr,
 					shared.EventReasonCluster,
-					fmt.Sprintf("Member Version is : %d. Was NIL", memberVersion),
+					"Member Version for {%s} is nil. Resetting.",
+					m.Pod,
 				)
+
+				memberVersion = connectionsVersion - 1
+
 			}
 
 			if memberVersion < connectionsVersion {
--- a/pkg/controller/kubedirectorcluster/members.go
+++ b/pkg/controller/kubedirectorcluster/members.go
@@ -284,7 +284,21 @@ func handleReadyMembers(
 				return
 			}
 
-			memberVersion := *m.StateDetail.LastConnectionVersion
+			var memberVersion int64
+
+			if m.StateDetail.LastConnectionVersion != nil {
+				memberVersion = *m.StateDetail.LastConnectionVersion
+
+			} else {
+				memberVersion = connectionsVersion - 1
+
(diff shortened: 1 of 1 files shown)


pkg/controller/kubedirectorcluster/members.go
```diff
@@ -284,7 +284,21 @@ func handleReadyMembers(
...
+				memberVersion = *m.StateDetail.LastConnectionVersion
+
+			} else {
+				memberVersion = connectionsVersion - 1
+
+				shared.LogInfo(
+					reqLogger,
+					cr,
+					shared.EventReasonCluster,
+					fmt.Sprintf("Member Version is : %d. Was NIL", memberVersion),
```
joel-bluedata, line 299: I'm not sure this is a useful message? If this is just a debug log message that needs to be removed, you can skip the rest of this comment.

However if it's something you think we should keep, a few tweaks are needed:
- If you just want to log this (and not generate an event), you can use shared.EventReasonNoEvent.
- Rather than building the message string yourself with fmt.Sprintf, use `shared.LogInfof` (rather than `shared.LogInfo` and pass the variables-to-be-substituted in as the final arguments. You can find examples of that elsewhere in this file.
- "Member Version" is too generic; we're specifically talking about "lastConnectionVersion" here.
- To really be a useful message we'd need the member name (from m.Pod) to be logged.

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -284,7 +284,21 @@ func handleReadyMembers(
...
 			}
 
-			memberVersion := *m.StateDetail.LastConnectionVersion
+			var memberVersion int64
+
+			if m.StateDetail.LastConnectionVersion != nil {
+				memberVersion = *m.StateDetail.LastConnectionVersion
+
+			} else {
+				memberVersion = connectionsVersion - 1
```
joel-bluedata, line 293: Can you tell me a bit more about why we default to "connectionsVersion - 1" if there's no lastConnectionVersion?

saurabh-jogalekar: Yes, I'm going to add comments and then convert this to a PR that is ready for review

joel-bluedata: 👍

pkg/controller/kubedirectorcluster/members.go
saurabh-jogalekar: Thanks, Joel. I've changed this to not generate an event.

pkg/controller/kubedirectorcluster/members.go
saurabh-jogalekar: Added comments in the PR description and code. Feel free to take a look.

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -284,7 +284,27 @@ func handleReadyMembers(
...
+			if m.StateDetail.LastConnectionVersion != nil {
+				memberVersion = *m.StateDetail.LastConnectionVersion
+
+			} else {
+
+				shared.LogInfof(
+					reqLogger,
+					cr,
+					"",
+					"Member Version for {%s} is nil. Resetting.",
```
joel-bluedata, line 301: Can we log "lastConnectionVersion" here instead of "Member Version"?

saurabh-jogalekar: Done

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -284,7 +284,27 @@ func handleReadyMembers(
...
+
+			if m.StateDetail.LastConnectionVersion != nil {
+				memberVersion = *m.StateDetail.LastConnectionVersion
+
+			} else {
+
+				shared.LogInfof(
+					reqLogger,
+					cr,
+					"",
```
joel-bluedata, line 300: Let's use the shared.EventReasonNoEvent constant here, rather than hardcoding its value.

joel-bluedata (commented):
👍 Generally looks good to me if Kartik is also OK w/ it.

kmathur2:
I am good too. Looks good.

kmathur2 (approved):

pkg/controller/kubedirectorcluster/members.go
saurabh-jogalekar: Done

saurabh-jogalekar:
Following scenarios are tested:
1. (v0.6.0) Scaling down replicas to 0 and scaling them back to 1+
2. (v0.6.0) Manually removing `lastConnectionVersion` from member status
3. KD Upgrade from v0.5.1 to this branch. Steps are: deploy KD 0.5.1 and create some kdcluster. Then delete the KD deployment resource, role, rolebinding, and service account. Then move over to a current KD codebase and apply all its CRDs, its rbac.yaml, and its deployment-localbuilt.yaml to get the new KD version coming up.

joel-bluedata (approved):

Issue #385, opened 2020-07-18
encrypted secrets


joel-bluedata:
@koshiksamota10 and @kmathur2 you will want to be in the loop for this one.

Cf. the discussion of secrets at the top of the [KubeDirectorCluster page in the wiki](https://github.com/bluek8s/kubedirector/wiki/KubeDirectorCluster-Definition). We should make it easier to support having encrypted content in a secret (or in a configmap, for that matter) by providing a key/password that is visible within the container (by app setup scripts) but not visible elsewhere.

I'll do some strawmanning/brainstorming in comments below.

joel-bluedata:
Let's say the secret content has been encrypted like this, by whoever then created the secret:

    openssl enc -aes-256-cbc -md sha256 -base64 -salt -in content.txt -out content.txt.enc

(That will interactively ask for a password.) Then when the setup script reads out the secret contents, it could decode it like this:

    openssl enc -aes-256-cbc -d -md sha256 -base64 -in content.txt.enc -out content.txt -pass pass:password_goes_here

One stake to put in the ground I think is that KubeDirector itself will not mandate an encryption method; some method would be mandated/documented for a particular purpose in a particular app, and the secret producer would need to use that method in order for the content to correctly get consumed by that app. Our responsibility as KD would just be to securely provide the key needed to decrypt the secret's contents.

First question is, where is this key specified. It possibly could be a per-app thing but I think our immediate usecases are per-cluster. I.e. the user or client creating the kdcluster knows what the key is and needs to provide it.

The naive way to get this key into the appropriate role members in the container would be to allow specifying a per-role key in the spec. We probably could and should even support a set of multiple keys, each marked with a different name/ID, and then the various secret mounts or connections could be configured as to which key should be used to decrypt their contents. Great! But then the keys are visible in the kdcluster so that's no improvement in security.

So, one possible way to address that key visibility:

* Have a master encryption key in the kd-global-config. Or even in some separate config CR with a different name, or a separate secret, just in case someone wants to put different privileges on it compared to kd-global-config. In any case this resource and the master encryption key it stores will be visible to KD and to cluster administrators but not to normal users.

* When a kdcluster is created with some keys specified, the KD validator webhook immediately takes those keys out of the spec, encrypts them with the master key, and stores them in the kdcluster's status stanza. (Or alternately we could store them back into the spec?)

* So only the encrypted keys will be visible in the kdcluster. This is useless to normal users, but useful for KD.

[...]

PR #481, opened 2021-05-18
#385 Secret Key implementation

aleshchynskyi:
Secret Keys are key-value pairs which can be set per each role. After submitting Cluster CR to validator, values are encrypted with master encryption key which is configurable in KDConfig. Decrypted values are only accessible within container as part of configmeta.json


pkg/validator/cluster.go
```diff
@@ -776,6 +781,57 @@ func validateSecrets(
...
+		var needsUpdate bool
+		for keyIndex, secretKey := range role.SecretKeys {
+			prevSecretKey := findSecretKey(prevCr, role.Name, secretKey.Name)
+			encryptedSecretKey := secretKey
+			if prevSecretKey != nil && secretkeys.Equal(*prevSecretKey, secretKey) {
+				encryptedSecretKey = *prevSecretKey
+			}
+			encryptedSecretKey, err := secretkeys.EncryptSecretKey(encryptedSecretKey)
+			if err != nil {
+				// TODO Better error handling
```
Vellanci, line 796: I've been thinking if it should log encryption error to KD logs

pkg/shared/types.go
```diff
@@ -36,6 +36,11 @@ const (
 	// DefaultNamingScheme - default naming scheme if not specified in
 	// the configCR
 	DefaultNamingScheme = "UID"
+
+	// DefaultMasterEncryptionKey - default master encryption key
+	// if not specified in configCR
+	// Generated 32 byte key for AES-256
+	DefaultMasterEncryptionKey = "7fc051adc074d25134d1987b978046fb77334c889e1997de3136854884241b9f"
```
Vellanci, line 43: Currently it's supposed to conform AES key format. We can either add validation for it in OpenAPI (CRD) or allow to use any passphrase and transform it into encryption key by some hashing algorithm.

pkg/validator/cluster.go
```diff
@@ -776,6 +781,57 @@ func validateSecrets(
...
 }
 
+// encryptSecretKeys encrypts secret keys per each role and generates patches if needed
+func encryptSecretKeys(cr *kdv1.KubeDirectorCluster, prevCr *kdv1.KubeDirectorCluster, valErrors []string, patches []clusterPatchSpec) ([]string, []clusterPatchSpec) {
+	for roleIndex, role := range cr.Spec.Roles {
+		var needsUpdate bool
+		for keyIndex, secretKey := range role.SecretKeys {
+			prevSecretKey := findSecretKey(prevCr, role.Name, secretKey.Name)
+			encryptedSecretKey := secretKey
+			if prevSecretKey != nil && secretkeys.Equal(*prevSecretKey, secretKey) {
```
Vellanci, line 791: It's my workaround for validation (and probably excessive configmeta changes). If secret key was already encrypted, reuse its encrypted value. The only problem is that it allows to bruteforce real value. But since it's made for encryption keys, it's basically impossible due to its length.

pkg/catalog/configmeta.go
```diff
@@ -327,12 +332,7 @@ func genClusterConnections(
 // representation. Note that KubeDirector currently only allows/manages one
 // nodegroup per virtual cluster, so this will always be a map that has a
 // single key of "1".
-func nodegroups(
-	cr *kdv1.KubeDirectorCluster,
-	appCR *kdv1.KubeDirectorApp,
-	membersForRole map[string][]*kdv1.MemberStatus,
-	domain string,
-) map[string]nodegroup {
+func nodegroups(cr *kdv1.KubeDirectorCluster, appCR *kdv1.KubeDirectorApp, membersForRole map[string][]*kdv1.MemberStatus, domain string) (map[string]nodegroup, error) {
```
joel-bluedata, line 335: Break up this long line here, using the function signature format above.

pkg/secretkeys/secretkeys.go
```diff
@@ -0,0 +1,106 @@
+package secretkeys
```
joel-bluedata, line 1: I'll come back to this file after I've looked at the wider logic.

One note, probably need some tweaks in this file (and elsewhere in the changes) to match to code format in the rest of KD. Particularly function signatures; if there's at least one function argument then they are of the form:
```
func myFuncName(
    arg1,
    arg2,
) returnType {
```
and all function signatures (whether zero or more args) have a blank line between the returnType and the start of the function body.

pkg/shared/types.go
joel-bluedata: IMO we shouldn't have any hardcoded default for the master key. If KD sees that there is no master key in its config, it should generate a random one and store it in the config.

pkg/validator/cluster.go
joel-bluedata: Might as well yep. As long as we don't log any actual keys.

pkg/validator/cluster.go
joel-bluedata: OK, hmm. It seems like we would just want to do:

* If value is nonempty, regardless of encryptedValue, we should calculate a new encryptedValue and set value to empty. (This means we'll need to generate a patch.)

* If value is empty, ignore this one.

pkg/validator/cluster.go
Vellanci: Yeah, I've been struggling on how to obtain logger in that function to do that

pkg/shared/types.go
Vellanci: Oh, it makes sense actually. It's much safer

pkg/validator/cluster.go
Vellanci: I just realized there cannot be excessive configmeta changes since decrtypted value doesn't change if we generate encrypted value for existing one. So your way will work just fine

pkg/secretkeys/secretkeys.go
Vellanci: Thanks for pointing it out, I didn't notice it at all.

kmathur2:
Overall changes looks good but I have a general question - 

If the decrypted key is available in configmeta.json then anyone with kubectl exec can decrypt secret contents right? How is this very different from base64 encoding?

kmathur2:
Also, just the decryption key won't be sufficient for apps to decrypt. Looking at the original strawman, the decrypt command example -
_openssl enc -aes-256-cbc -d -md sha256 -base64 -in content.txt.enc -out content.txt -pass pass:password_goes_here_

Secret creator/encryptor and app developers could be entirely different personas so app devs would probably need more to implement the decryption logic than just the decryption keys. 

@joel-bluedata thoughts?

aleshchynskyi:
> 
> 
> Overall changes looks good but I have a general question -
> 
> If the decrypted key is available in configmeta.json then anyone with kubectl exec can decrypt secret contents right? How is this very different from base64 encoding?

Basically I think the range of roles is narrowing. As you said a person with exec rights is able to see the secret key's content, but usual secret contents is available to anyone with view rights or to anyone who can mount secret to a newly created pod.

That being said it's still possible to create a simple KDCluster (ubuntu for example) with the same secret key and see it in its configmeta.json.

aleshchynskyi:
> 
> 
> Also, just the decryption key won't be sufficient for apps to decrypt. Looking at the original strawman, the decrypt command example -
> _openssl enc -aes-256-cbc -d -md sha256 -base64 -in content.txt.enc -out content.txt -pass pass:password_goes_here_
> 
> Secret creator/encryptor and app developers could be entirely different personas so app devs would probably need more to implement the decryption logic than just the decryption keys.
> 
> @joel-bluedata thoughts?

I think as Joel mentioned in the issue, encrypted method is mandated by app developer. So it should be known to secret creator that app is using aes-256-cbc with sha256 passphrase digest and base64 encoding. KD should only provide a way to store decryption key. Also, if it's needed, they can parametrize decryption info in env vars or secret keys (if they consider that info sensitive).

joel-bluedata:
> That being said it's still possible to create a simple KDCluster (ubuntu for example) with the same secret key and see it in its configmeta.json. It can be a reason why we should move encrypted value to status (to forbid placing already encrypted values). Or even remove it from custom resource at all (if its possible).

Yeah I think we have three choices here:

* Have the webhook handler put the encrypted user key in status rather than spec, so it is automatically protected from anyone else changing it.
* Have the webhook handler block any incoming change to the encrypted key.
* Include a salt in the encrypted key which differs per-kdcluster or even per-role. However this salt would then need to be stored in a way that the user can't change it, which really comes back to the above two kinds of solutions.

We can't remove the encrypted key from the custom resource as long as we're treating the CR as the place where we keep our persistent state.

So I do think the encrypted key either needs to go into the status stanza, or if it stays in the spec then the validator needs to reject any incoming changes for it.

joel-bluedata:
Also coming back to the idea of salt for a minute though: if we _did_ have a different random salt per user key, that would prevent any outside viewer from seeing when two keys are the same. Which would be beneficial.

joel-bluedata:
> Basically I think the range of roles is narrowing. As you said a person with exec rights is able to see the secret key's content, but usual secret contents is available to anyone with view rights or to anyone who can mount secret to a newly created pod.

Yeah this is the idea. If someone can get inside the pod, they can already see whatever the contents of the secret has been used for; e.g. if it is an LDAP authentication password they can see the password in the pod's LDAP configuration files. (Modulo considerations about different user privileges inside the pod, but I don't think we can assume that the in-pod services and files are going to be under privileged accounts that are hidden from the pod exec user.) So I don't think it's an additional leak that they could also see the key itself in that case. pod exec privileges should be pretty tightly controlled.

As Anton says we just want to keep the secret contents from leaking to people who have other privs like the ability to see pods and create pods/kdclusters, fetch secrets by name, etc.

joel-bluedata:
> I think as Joel mentioned in the issue, encrypted method is mandated by app developer. So it should be known to secret creator that app is using aes-256-cbc with sha256 passphrase digest and base64 encoding. KD should only provide a way to store decryption key. Also, if it's needed, they can parametrize decryption info in env vars or secret keys (if they consider that info sensitive).

Yep. Part of the app usage instructions would need to be: "give me a secret that has its contents encoded in this way".

Before we merge this PR we should do a sanity check of that usage model with a current secret-consuming app.

aleshchynskyi:
> Also coming back to the idea of salt for a minute though: if we _did_ have a different random salt per user key, that would prevent any outside viewer from seeing when two keys are the same. Which would be beneficial.

There's already random nonce that does that. Every call of encrypt function returns different value. It's stored as first N bytes of encrypted value (depends on encryption key length I believe).

aleshchynskyi:
The idea of putting encrypted values to status messes up workflow a bit. When creating a KDCluster status field is yet empty, especially role statuses which is breaking the idea. It's created later in controller while encryption happens in admission webhook. 

The same deal with making an encryption salt per role.

So it leaves us with 2 ways:
1. Block user changes to encryptedValue and allow only deleting it.
2. Add an encryption salt per kdcluster.

aleshchynskyi:
Updated:
- Forbidden manual encrypted value placement
- masterEncryptionKey validation for 16,24,32 length
- masterEncryptionKey random default value
- Validator logging for encryption issues

Note: secretKeys cannot be changed due to KD role changes rejection logic

@kmathur2 @joel-bluedata please review

joel-bluedata:
Thanks! Will review. Is there a concise example we could document/share for this feature?

aleshchynskyi:
Found a possible problem: when KDConfig is not created by user, initial master encryption key generation won't work since it's based on webhook. Currently it will raise an error that master encryption key is required to Secret Keys to work.

kmathur2 (approved):

joel-bluedata (approved):
Alll righty! Good work, and thanks for your patience. I'll update the CRD wiki pages soon.

PR #483, opened 2021-05-19
Fix appPrepConfigRunCmd to report stderr and status

kmathur2:
Minor tweaks to appPrepConfigRunCmd to fix stderr and status which was getting lost.

Even a failed startscript was not writing anything to /opt/guesstconfig/configure.stderr or /opt/guesstconfig/configure.status.

Tested all /opt/guestconfig/configure.* for successful and unssucessful runs of startscript.

### DIFF
--- a/pkg/controller/kubedirectorcluster/types.go
+++ b/pkg/controller/kubedirectorcluster/types.go
@@ -89,10 +89,7 @@ const (
 	rm -rf /opt/guestconfig/appconfig.tgz`
 	appPrepConfigStatus = "/opt/guestconfig/configure.status"
 	appPrepConfigRunCmd = `rm -f /opt/guestconfig/configure.* &&
-	echo -n %s= > ` + appPrepConfigStatus + ` &&
-	nohup sh -c "` + appPrepStartscript + ` --configure;
-	echo -n $? >> ` + appPrepConfigStatus + `" > /opt/guestconfig/configure.stdout  
-	2> /opt/guestconfig/configure.stderr  &`
+	echo -n %s= > ` + appPrepConfigStatus + ` && nohup sh -c "` + appPrepStartscript + ` --configure 2>/opt/guestconfig/configure.stderr 1>/opt/guestconfig/configure.stdout;echo $? >> ` + appPrepConfigStatus + `" &`
 	fileInjectionCommand = `mkdir -p %s && cd %s &&
 	curl -L %s -o %s`
 	appPrepConfigReconnectCmd = `echo -n %s= > ` + appPrepConfigStatus + ` &&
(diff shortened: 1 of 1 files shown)


pkg/controller/kubedirectorcluster/types.go
```diff
@@ -89,10 +89,9 @@ const (
 	rm -rf /opt/guestconfig/appconfig.tgz`
 	appPrepConfigStatus = "/opt/guestconfig/configure.status"
 	appPrepConfigRunCmd = `rm -f /opt/guestconfig/configure.* &&
-	echo -n %s= > ` + appPrepConfigStatus + ` &&
-	nohup sh -c "` + appPrepStartscript + ` --configure;
-	echo -n $? >> ` + appPrepConfigStatus + `" > /opt/guestconfig/configure.stdout  
-	2> /opt/guestconfig/configure.stderr  &`
+	echo -n %s= > ` + appPrepConfigStatus + ` && 
+	nohup sh -c "` + appPrepStartscript + ` --configure 2>/opt/guestconfig/configure.stderr 1>/opt/guestconfig/configure.stdout;
+	echo $? >> ` + appPrepConfigStatus + `" &`
```
joel-bluedata, line 94: Was removing the "-n" option here intentional?

kmathur2: Yes, I think it's not required since line 92 writes 
```
docker://789aabbca73481a9b0b2f0891281cd32e05995353287a9e1b07810bcf6d1417c=0
```
which is what we need

joel-bluedata: Right, but previously we didn't have a trailing carriage return after all that. It may not matter, depends on the processing code, but just wondering why you put that final carriage return back in.

kmathur2: On line 93 ?
For readability, it's basically split into 3 subcommands 

line 92 - write container id in status
line 93 - run startscript in the background and write stdout and stderr in respective files
line 94 - update status file with exit code of the previous command.

joel-bluedata: Yeah, it's the change in line 94 that I'm asking about.

kmathur2: ok, I see this is the original question of why did we remove "-n". Actually it is not making any diff in status. I always see docker://789aabbca73481a9b0b2f0891281cd32e05995353287a9e1b07810bcf6d1417c=0
We can add -n back.

But there is a new problem that I found. With nohup the $? is always returning 0.
Trying to debug that

kmathur2: Added -n back, and replaced double quote with a single.

This now works as expected

neogeographica: Are we sure. :-)

Man this particular code has been ornery.

joel-bluedata: Are we sure. :-)

Man this particular code has been ornery.

joel-bluedata (approved):

PR #489, opened 2021-05-26
update cr to 2.1

ginochen:


joel-bluedata:
@ginochen should we close this or do something with it?

ginochen:
Yes, it's not needed

PR #490, opened 2021-05-29
API serving tag update

Supekar:


joel-bluedata:
@Supekar should we close this or do something with it?

PR #499, opened 2021-06-18
fix some docs-rot and add a roadmap

joel-bluedata:
Still open to roadmap changes; putting this PR up so that folks can see it.


kmathur2 (approved):

kmathur2:
👍

Issue #500, opened 2021-06-21
kdcluster backup/restore enablement


joel-bluedata:
Backup-and-restore for kdclusters currently has three issues.

- A kdcluster needs to be restored AFTER the relevant kdapp.

- Depending on how we attack this, the kdcluster probably also needs to be restored after the native K8s resources that make up the kdcluster.

- We lose the kdcluster status stanza, when that kdcluster resource is re-created. There's important stuff in there!

**Thoughts about ordering w.r.t. the kdapp:**

If we capture the specific distro ID and version of the kdapp in the kdcluster status, then we could do something like:
- If we detect that this is a restored kdcluster rather than a new one, don't immediately reject a kdcluster that has a missing kdapp.
- Once we're in a position to do reconciliation, if the kdapp is still missing then just post an Event describing the problem and skip reconciliation.
- If/when the kdapp finally appears, reconciliation can proceed normally.

(To detect when we are restoring a kdcluster, rather than creating a new one; e.g. we could check for the presence of KD's finalizer.)

It's important to track & honor the distro ID and version, to make sure we really are reconnecting with the same kdapp -- not some other kdapp resource created in the interim between backup/restore that happens to have the same name as the old one but different contents.

**Thoughts about ordering w.r.t. component K8s resources:**

This one is hard to handle programmatically, as from KD's perspective there's no detectable difference between "the restore process hasn't restored my statefulset yet" vs. "my statefulset is missing and must be re-created".

We seem to only have two choices. Either require that kdcluster gets restored last, or have a post-restoration kdcluster NOT immediately start reconciliation until it is explicitly told to do so. (The "explicit tell" could take many forms.)

**Thoughts about losing the status stanza:**

Ideally, perhaps, our important state would all be out there in the native K8s resources. Some thoughts about that (and why we aren't doing it currently) in this gist: https://gist.github.com/joel-bluedata/ce39dd74f960fe773ad20a011eb7086d

Another alternative to using the status stanza would be to use an actual database rather than storing stuff in etcd documents. This isn't currently under consideration, but we're aware that it's a possibility.

The track we're pursuing at the moment involves storing the state for a kdcluster in some other document. There would be some advantages to stashing these documents in their own special namespace, but that complicates some common backup/restore scenarios. So it looks like such a document would live in the same namespace as its kdcluster.

It would be somewhat natural to use a configmap or secret for this purpose. Drawbacks however:
[...]

PR #508, opened 2021-09-09
Backup/restore support

joel-bluedata:
closes issue #500 

Posting as a draft for starters while I do more testing and decide how to explain/doc it. (Also need to rebase to master.)

The individual commits are fairly self-contained. Might be a good idea to look at them instead of, or supplemental to, looking at the whole diff.

==========

Issue #500 covers the problem we're addressing here, but let me summarize.

KD has three challenges in a backup-and-restore situation:

1) The status stanza of the kdcluster is lost.

2) Owner references on various kdcluster-related components are lost (or made stale). Without these, deleting a kdcluster will not properly clean up everything.

3) KD can't make appropriate reconciliation decisions until all relevant resources are restored. For example if a statefulset for a role is missing, KD doesn't know if it's actually missing or if the backup solution is about to get around to restoring it.

There are lots of other issues in doing backup-and-restore properly, especially with complex resource arrangements e.g. statefulsets with pods that mount PVs, but these issues are not KD-specific. Any KD changes here assume that the backup-and-restore solution is handling all the "plain Kubernetes" issues appropriately.

We're specifically testing with Velero, so no assurances for any other solution, but we don't expect there to be big differences in what will be required of KD.

==========

The solution for challenge 1 is to (optionally) "back up" the status stanza for each kdcluster in a separate CR in the namespace. We're using a CR rather than e.g. a configmap because we want to make it easy to control its ACLs, and to not pollute any listings of configmaps. Really only KD should care about this resource.

There's a new KD config flag, backupClusterStatus, that defaults to false.

If it is set true, any write to a kdcluster status stanza will also write the same content to a kdstatusbackup CR of the same name as the kdcluster (creating that CR if necessary).

If it is set false, any existing kdstatusbackup CRs will be deleted.

Note that changing the backupClusterStatus value will only affect a kdstatusbackup CR on the next reconciliation handler for its kdcluster.

==========

The solution for challenge 2 is to include owner reference repair/enforcement in the reconciliation of all relevant resources.

==========

The solution for challenge 3 is somewhat more involved.

A kdcluster depends on 3 kinds of things: its kdapp CR, its kdstatusbackup CR, and any K8s-specific resources that are used to implement the kdcluster. For that last category, we really only care about the resources that KD will want to explicitly create/reconfigure/delete: per-role statefulsets, per-member services, and the cluster (headless) service. Anything created by the statefulset controller is a thing where KD already properly handles cases of things appearing/disappearing asynchronously to its control.

So if a kdcluster is restored, we want to wait for these things to appear before resuming reconciliation. In the case of the kdstatusbackup CR, when it reappears its contents can be copied back into the kdcluster's status stanza.

KD will detect a "restored" kdcluster by an annotation that was previously placed on the CR (if necessary) when the kdcluster status was updated. If this annotation is seen on a kdcluster that is being created, the validator skips validation of the incoming CR and sets the kubedirector.hpe.com/restoring label on the CR. (No label value.)

The presence of this label prevents the normal reconciliation from happening for this kdcluster. Instead, the handler will watch for those necessary resources (and set some status flags to let any observer know what it is still waiting on). If those resources all reappear, KD will remove the label and let reconciliation resume.

Various things to note about this process:

* While the restoring label is in place, the spec of the kdcluster cannot be modified.

* A non-KD client can also manually remove the restoring label, to force reconciliation to resume. However any attempt to remove the label (including by KD) will trigger validation of the kdcluster, and if validation fails the label cannot be removed.

* While the restoring label is in place, the kdcluster normally cannot be deleted. If you were allowed to delete a kdcluster while its resources were still being restored, you might end up with orphaned resources floating around in your namespace.

* However. If you absolutely do need to delete a kdcluster, and you don't want to (or can't, because of validation failures) remove the restoring label, then you can first add the kubedirector.hpe.com/allow-delete-while-restoring label to the CR (label value does not matter). Once this label is in place, deleting the kdcluster will be allowed even if it has the restoring label.


joel-bluedata:
I did a fairly late switch from having an explicit "this CR was previously created" annotation to instead relying on our finalizer. While this worked fine in manual "restore" tests, now that I've jumped back over to using Velero it looks like maybe Velero strips the finalizers on restore? Need to investigate that a bit more but probably I'll need to go back to using an annotation.

joel-bluedata:
Note that the doc text assumes that PR #509 has also been merged.

joel-bluedata:
Found a bug with the owner-ref repairing. The statefulset not only has its own owner-ref, but also an owner-ref in the volumeClaimTemplates. This means that after the sset is restored, any newly created PVCs will be automatically deleted, which is pretty puzzling when you see it in action.

I'm kinda surprised Velero left that reference in there? Well in any case, even if they had stripped it out, we'd still need to repair/restore it. (Otherwise we'd get the opposite undesirable behavior of PVC remaining after sset deletion.)

joel-bluedata:
Hmm. We're not allowed to edit that part of the spec in an existing statefulset.

There's no other way to indicate that PVCs should be auto-deleted, at least not until https://github.com/kubernetes/enhancements/pull/2440 arrives in some future K8s version.

I guess we could make the KD webhook fix up the owner ref on any PVCs created with the kubedirector.hpe.com/kdcluster label. Kind of annoying, but should work... it should even fix PVCs from statefulsets created by previous versions of KD and then restored, which have this problem of an incorrect owner ref.

deploy/kubedirector/kubedirector.hpe.com_kubedirectorstatusbackups_crd.yaml
```diff
@@ -0,0 +1,31 @@
...
+        apiVersion:
+          type: string
+        kind:
+          type: string
+        metadata:
+          type: object
+        spec:
+          type: object
+          properties:
+            statusBackup:
```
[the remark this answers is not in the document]
joel-bluedata: In practice it is, in that the "object" value here is the complete kdcluster status stanza.

Ideally (arguably) the validation schema here should have all of the stuff that is in the validation schema for the status stanza. However I really don't want to manually copy all of that; that seems like a recipe for disaster when future status changes happen.

In the end since it is only KD that is writing/reading this -- and only by copying directly from and writing directly to the kdcluster status stanza  -- I'm OK with not having a validation schema here.

If we did want to have a schema here, we could for example come up with a way to define the status schema only in one document and then have the "make deploy" process build the actual CRDs used for kdcluster and kdstatusbackup. However that would introduce problems for folks who just want to directly apply the CRDs out of this repo for development/testing, upgrade, or whatever.

pkg/validator/pvc.go
```diff
@@ -0,0 +1,102 @@
...
+// pvcPatchSpec is used to create the PATCH operation for setting the PVC
+// owner references.
+type pvcPatchSpec struct {
+	Op    string                  `json:"op"`
+	Path  string                  `json:"path"`
+	Value []metav1.OwnerReference `json:"value"`
+}
+
+// mutateOwnerRefs checks PVC labels to see if it is for a kdcluster. If it
+// is, it returns patches to set the ownerReferences to point to that CR.
```
kmathur2, line 24: 👍

kmathur2: what about owner ref to statefulset, service etc?

joel-bluedata: We fix those in reconcile instead; see UpdateStatefulSetNonReplicas, UpdateHeadlessService, and UpdatePodService.

Since those owner references are editable (and since we're already in the business of reconciling them) we can do that rather than needing to introduce more webhooks for statefulsets and services.

joel-bluedata: There is kind of an asymmetry here which I'll say a bit more about.

One of the reasons to have a webhook here -- rather than introducing reconciliation handling for PVCs -- is to make sure we handle restored statefulsets that were created by an old version of KD, with the PVC ownerref included in the PVC template of the statefulset. If we don't fix those PVCs by intervening in their creation, then as soon as they are created with an old/stale owner ref they will vanish.

Since we have to have the webhook to handle PVCs generated by those old statefulsets, and the webhook also handles the "normal" case of PVC creation by current-form statefulsets, then we don't need to do any PVC reconciliation.

I guess that does mean that if someone comes along later and manually edits the owner ref on the PVC to remove the kdcluster as owner, we don't forcibly re-instate our kdcluster owner ref there. Might be something we want to think about doing but at least that's not a regression -- that's the way it's always been. I chose not to write a whole new PVC reconciler in this PR just for that purpose. :-)

deploy/kubedirector/kubedirector.hpe.com_kubedirectorstatusbackups_crd.yaml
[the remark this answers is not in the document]
kmathur2: sure make sense, if someone manually tries to tinker with backup they can easily put KD in panic state. Is that ok?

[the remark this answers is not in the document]
kmathur2: In the Tenant Role/RoleBinding, we should revoke all access for everyone for kdstatusbackup (Not in this PR obviously)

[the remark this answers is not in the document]
joel-bluedata: Yeah, point taken. Basically nobody but KD and the backup solution should be given privileges on kdstatusbackup CRs (one of the reasons we're using a custom CR rather than say ConfigMap).

If we did have a validation schema here it would help somewhat to protect against unexpected manual fiddling, although I think probably you could still also construct something that would upset KD.

doc/backup-and-restore.md
```diff
@@ -0,0 +1,96 @@
...
+    restoreProgress:
+      awaitingApp: false
+      awaitingResources: false
+      awaitingStatus: false
+      error: ""
+```
+
+The "awaitingApp" flag will switch to true if the relevant kdapp is restored. The "awaitingStatus" flag becomes true when the relevant kdstatusbackup is restored. Once status is restored, the "awaitingResources" flag can become true if all resources named in the status are restored. Once this happens, KubeDirector will automatically resume reconciliation for this kdcluster.
+
+If/when reconciliation resumes, the kdcluster will be validated, running through the same series of checks that would be normally be used when the kdcluster is initially created. If this validation fails then reconciliation cannot resume, and the "error" field of "restoreProgress" will describe the validation failure.
```
kmathur2, line 73: nit: that would ~~be~~ normally be ...

pkg/apis/kubedirector/v1beta1/kubedirectorstatusbackup_types.go
```diff
@@ -0,0 +1,50 @@
+// Copyright 2019 Hewlett Packard Enterprise Development LP
```
kmathur2, line 1: nit:2021 :)

pkg/controller/kubedirectorcluster/cluster_restore.go
```diff
@@ -0,0 +1,411 @@
...
+		}
+	}
+	return true
+}
+
+// roleResourcesExist looks to see if a statefulsets named in the
+// status exist, along with the necessary per-member services.
+func roleResourcesExist(
+	reqLogger logr.Logger,
+	cr *kdv1.KubeDirectorCluster,
```
kmathur2, line 320: what about _connection_ resources? Like configmap, secret or another kdcluster?  what happens if it was connected to another kdcluster and that is not yet up ?

pkg/controller/kubedirectorcluster/cluster_restore.go
```diff
@@ -0,0 +1,411 @@
...
+// checkRestoreDone will try to clear the being-restored label if all the
+// "waiting" flags are false. If that fails it will populate the error field
+// of the restore progress object. Return true if and only if we have
+// successfully cleared that label.
+func checkRestoreDone(
+	reqLogger logr.Logger,
+	cr *kdv1.KubeDirectorCluster,
+) bool {
+
+	if cr.Status.RestoreProgress.AwaitingApp ||
```
kmathur2, line 372: can we come here after marking error in last run? should we check error also here?

pkg/controller/kubedirectorcluster/cluster_restore.go
```diff
@@ -0,0 +1,411 @@
...
+	// Don't want to use LogError here because it generates a stacktrace.
+	// We don't really care about that, just about the message.
+	shared.LogInfof(
+		reqLogger,
+		cr,
+		shared.EventReasonCluster,
+		"failed to resume reconciliation: %s",
+		patchErr.Error(),
+	)
+	cr.Status.RestoreProgress.Error = patchErr.Error()
```
kmathur2, line 409: shouldn't all the flags be marked false here?

kmathur2:
Do we need to run --configure after kdcluster is restored?

pkg/controller/kubedirectorcluster/cluster_restore.go
joel-bluedata: Yeah that's a good question. Initially you'd have the same situation as if the configmap (or kdcluster or whatever) had not been created yet, i.e. it would be omitted from the configmeta sent into the container. Then when the object comes back, our connections hash will change, we'll modify the configmeta to include it, and do a reconnect notify.

If an app doesn't support the reconnect notify and always expects its connections to be there before it gets itself configured, then it could be upset at that. Do you think we need the semantic where reconciliation is blocked until connections come back? A potential kicker with that, is that there might have been a missing connection to begin with (at the time the kdcluster was backed up) and so that one will _never_ "come back" on the restore.

joel-bluedata:
> Do we need to run --configure after kdcluster is restored?

That's handled by the normal "container has changed" logic already. When the container changes, if it's using persistent storage we don't re-run. If it's not persistent, we do.

pkg/controller/kubedirectorcluster/cluster_restore.go
joel-bluedata: The flags' true-ness still represents the known state of things, we just failed to automatically resume reconciliation.

pkg/controller/kubedirectorcluster/cluster_restore.go
joel-bluedata: I think even if there was previously an error, it's ok to let KD try again (on next handler) to auto-resume reconciliation. It will basically keep trying until it succeeds, or some of the dependencies go back to false again, or the kdcluster is deleted.

There are probably only two broad scenarios where it might be likely to fail an attempt and then succeed on a subsequent attempt:

* Temporary failure to interact with the API server for whatever reason.

* Validation failure that can be humanly resolved. It's tough to imagine many cases for this one, especially if you're just doing "back up everything" and then "restore everything" cases for a namespace. However if you were trying to do something more fiddly, like only back up kdclusters and manually manage the presence/absence of kdapps, then you might get in a situation where the kdapp with the name referenced by the kdcluster is different from the one that was being used at backup time. In that (contrived) case, validation will fail until you get the correct kdapp back in there.

pkg/controller/kubedirectorcluster/cluster_restore.go
kmathur2: Hmmm, currently most of the apps do not implement --reconnect since it was added later. Hence it will be a common scenario which we should worry about. So, yes blocking restore until all connections are not up will be cleaner. Can we check for missing connection at the backup time also and if there is then we store that info somewhere. 
This can be very involved and we can do this in another PR but someone like GL would definitely need this if they want to utilize backup functionality for existing apps like notebook with connections.

kmathur2: On similar lines, will sequencing matter for these features - 
1. blockStorageDevice
2. serviceAccountName
3. secret mount

pkg/controller/kubedirectorcluster/cluster_restore.go
kmathur2: ok makes sense. We are okay here then

kmathur2:
Just a few questions, overall looking great!!

Will start testing now

pkg/controller/kubedirectorcluster/cluster_restore.go
joel-bluedata: In general for things like that we shouldn't be concerned. Those are items that went into constructing the pod template inside the statefulset spec, and our philosophy about backup/restore should be that making the native K8s objects work correctly after restoration is up to the backup/restore solution to handle.

For example if an indicated service account is missing, the statefulset controller will fail to create a pod until the service account exists. If a secret-to-mount is missing, the sset controller can create the pod but it won't get out of ContainerCreating state until the secret exists. In those cases ordering doesn't really matter a lot as the pod container won't finally come up until those things are present.

For the block storage, similar to the other filesystem-mode persistent storage, Velero should recreate the PVs before recreating the statefulset (with appropriate claimRefs pointing back to the PVCs that the statefulset controller will create). I'm not sure what other backup solutions will do but I guess as mentioned above this is the sort of thing I would put "below the line" of KD's concern. I.e. anything that would have to be taken care of to make a "normal" statefulset work, we can depend on for the statefulsets that make up KD roles.

joel-bluedata:
I pushed the code change for wait-for-connections so you can see it in this PR (commit 2d2e48dc83c6711b2bd218e3d168e9b303fa0aa4). Not tested yet; I'll test it this evening once you're done with our test cluster.

kmathur2:
> I pushed the code change for wait-for-connections so you can see it in this PR (commit [2d2e48d](https://github.com/bluek8s/kubedirector/commit/2d2e48dc83c6711b2bd218e3d168e9b303fa0aa4)). Not tested yet; I'll test it this evening once you're done with our test cluster.

Sounds good. Feel free to use the system in the eve. I have to be away in the afternoon. Will resume testing late evening after you are done.

kmathur2 (approved):

kmathur2:
🚢  🚢

joel-bluedata:
OK, away we go!

PR #509, opened 2021-09-10
More control over labels and annotations

joel-bluedata:
Allow user-specified annotation in the same way we allow user-specified labels.

Also, let the KD global config specify labels/annotations that will be applied to any pod or member service.

A specific use case for this is to control Velero backups, where we NEVER want Velero to try to back up the tmpfs volumes of KD-created pods. With this change, you can use the global config to specify the following pod annotation to achieve that: backup.velero.io/backup-volumes-excludes: tmpfs-tmp,tmpfs-run,tmpfs-run-lock


kmathur2:
Looks good!

kmathur2 (approved):

Issue #475, opened 2021-05-14
Support setting up serviceAccounts for role


swamibluedata:
To support applications that will have to interact with api server, it is essential to support serviceaccount injection to all member of a role. User would have created the service account prior to this. (For ECP created tenant, we have the serviceAccount automatically created).

Can we support this through cluster cr definition?

kmathur2:
We have connections that currently takes (configmap, secret, cluster) maybe we can add service account to that.

joel-bluedata:
There's an interesting twist/variant for this that we've been discussing elsewhere. If the kdapp is an operator (or similar thing), it may need very wide privileges for its service account... you may in fact not want to have a service account with those privs existing within the same namespace as the kdcluster. I.e. your cluster admin may have decided it's OK for you (working within this namespace) to deploy this specific kdapp that has wide privileges, but not for you to be able to create/deploy other kdapps or pods etc. that have the same set of privs.

We need some sort of feature where a cluster admin can "bless" certain roles for certain kdapp distro IDs, to give them the ability to do k8s API requests with certain privileges from inside their containers. Probably need a kdapp checksum there as well to make sure it's the one we think it is.

This "blessing" could be done by doing things in the KD namespace, for example editing the KD config to list the role/distroID/checksums in question.

The privileges (Role resource content) probably need to be described by the kdapp, and then KD would do what to get those privs usable from within the container? Since we can't assume any particular kind of external authn/authz support, it seems like service accounts are what we have to work with. I need to refresh my understanding of the kinds of shenanigans we could do here, but for example maybe KD would: create the relevant role in the kdcluster's namespace, create a service account in the *KD* namespace, bind that SA to that role. Then have KD explicitly copy the SA token into appropriate containers when needed.

PR #510, opened 2021-09-10
Ability to provide Service Account at KDCluster level

kmathur2:
Fix https://github.com/bluek8s/kubedirector/issues/475

### DIFF
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
@@ -147,6 +147,9 @@ spec:
                           ephemeral-storage:
                             type: string
                             pattern: '^([0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
+                      serviceAccountName:
+                        type: string
+                        pattern: '^([0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
                   env:
                     type: array
                     items:
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
@@ -147,9 +147,9 @@ spec:
                           ephemeral-storage:
                             type: string
                             pattern: '^([0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
-                      serviceAccountName:
+                  serviceAccountName:
                         type: string
-                        pattern: '^([0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
+                        minLength: 1
                   env:
                     type: array
                     items:
--- a/pkg/executor/statefulset.go
+++ b/pkg/executor/statefulset.go
@@ -252,7 +252,10 @@ func getStatefulset(
 		}
 	}
 
-	useServiceAccount := true
+	useServiceAccount := false
+	if role.ServiceAccountName != "" {
+		useServiceAccount = true
+	}
 	volumeMounts, volumes, volumesErr := generateVolumeMounts(
 		cr,
 		role,
--- a/pkg/executor/statefulset.go
+++ b/pkg/executor/statefulset.go
@@ -340,7 +340,7 @@ func getStatefulset(
 						persistDirs,
 					),
 					Affinity:           role.Affinity,
-					ServiceAccountName: *role.ServiceAccountName,
+					ServiceAccountName: role.ServiceAccountName,
 					Containers: []v1.Container{
 						{
 							Name:            AppContainerName,
--- a/pkg/executor/statefulset.go
+++ b/pkg/executor/statefulset.go
@@ -252,7 +252,7 @@ func getStatefulset(
 		}
 	}
 
-	useServiceAccount := false
+	useServiceAccount := true
 	volumeMounts, volumes, volumesErr := generateVolumeMounts(
 		cr,
 		role,
@@ -306,7 +306,6 @@ func getStatefulset(
 	}
 
 	vct := getVolumeClaimTemplate(cr, role, PvcNamePrefix)
-
 	return &appsv1.StatefulSet{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "StatefulSet",
@@ -340,7 +339,8 @@ func getStatefulset(
 						imageID,
 						persistDirs,
 					),
-					Affinity: role.Affinity,
+					Affinity:           role.Affinity,
+					ServiceAccountName: *role.ServiceAccountName,
 					Containers: []v1.Container{
 						{
 							Name:            AppContainerName,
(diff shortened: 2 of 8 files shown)


kmathur2:
@joel-bluedata @swamibluedata should we support service accounts by labels also along with names? This could be really handy in the ECP world where apps would like to consume hpecp-tenant-* sa. Thoughts?

deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
```diff
@@ -147,6 +147,9 @@ spec:
                           ephemeral-storage:
                             type: string
                             pattern: '^([0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
+                      serviceAccountName:
```
joel-bluedata, line 150: I think this is in the wrong place... you've got it as part of "resources" rather than part of the role spec.

pkg/executor/statefulset.go
```diff
@@ -252,7 +252,7 @@ func getStatefulset(
 		}
 	}
 
-	useServiceAccount := false
+	useServiceAccount := true
```
joel-bluedata, line 255: What if serviceAccountName is nil or empty?

pkg/executor/statefulset.go
```diff
@@ -340,7 +339,8 @@ func getStatefulset(
 						imageID,
 						persistDirs,
 					),
-					Affinity: role.Affinity,
+					Affinity:           role.Affinity,
+					ServiceAccountName: *role.ServiceAccountName,
```
joel-bluedata, line 343: same question

pkg/executor/statefulset.go
kmathur2: k8s allows automountServiceAccountToken to be true without serviceAccountName. So we shouldn't need explicit validation.

swamibluedata:
Don't we want to treat serviceaccounts as secrets? How can we protect a user from using another user's service account?

deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
```diff
@@ -147,6 +147,9 @@ spec:
                           ephemeral-storage:
                             type: string
                             pattern: '^([0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
+                  serviceAccountName:
+                        type: string
+                        minLength: 1
```
swamibluedata, line 152: We need to validation for this to make sure the serviceaccount exists in that namespace

swamibluedata:
> Don't we want to treat serviceaccounts as secrets? How can we protect a user from using another user's service account?

Maybe a kdconfig parameter to make this configurable?

pkg/executor/statefulset.go
swamibluedata: Kubernetes will mount the default service account if automountServiceAccountToken is set (which is also the default) and serviceAccountName is not specified. That may not be the intended behavior. Why surface the service account if not requested. We should continue to set automountServiceAccountToken to false if service account is not provided through cluster cr

kmathur2: Hmm, yeah makes sense.

kmathur2:
> > Don't we want to treat serviceaccounts as secrets? How can we protect a user from using another user's service account?
> 
> Maybe a kdconfig parameter to make this configurable?

If the user knows another user's SA they can always spawn a pod or statefulset with that, how is this any different?

joel-bluedata:
> If the user knows another user's SA they can always spawn a pod or statefulset with that, how is this any different?

Yeah even if we hide/encrypt the SA name in the kdcluster ... if they can see the pod that is generated by KD, there's nothing we can do about that.

I can think of ways to give a KD cluster some credentials that would be protected from anyone "snooping" the kdcluster and its pods -- one way is described in the discussion for issue #475, and there are other things we could do using the new encrypted-secrets feature. But those are quite a bit more work. My understanding was that the complicated features were being pushed for now, in favor of a feature like this which just allows you to request the existing K8s behavior.


> @joel-bluedata @swamibluedata should we support service accounts by labels also along with names? This could be really handy in the ECP world where apps would like to consume hpecp-tenant-* sa. Thoughts?

Yes it does seem like it would be nice if we could accept a selector, but what should we do if that matches multiple SAs?

swamibluedata:
> > > Don't we want to treat serviceaccounts as secrets? How can we protect a user from using another user's service account?
> > 
> > 
> > Maybe a kdconfig parameter to make this configurable?
> 
> If the user knows another user's SA they can always spawn a pod or statefulset with that, how is this any different?

Here is the issue though. KubeDirector has cluster role which will bypass any user specific role. It is possible to limit a user to consume a particular serviceaccount with RBAC. (by using resourceNames)

swamibluedata:
> I dunno... @swamibluedata did you have something in mind?

Can we somehow use SubjectAccessReview for this to make sure the user is authorized to access the specified serviceAccount. userInfo in admissionRequest should have the information needed?

kmathur2:
> > I dunno... @swamibluedata did you have something in mind?
> 
> Can we somehow use SubjectAccessReview for this to make sure the user is authorized to access the specified serviceAccount. userInfo in admissionRequest should have the information needed?

Hmmm, so for this PR we need to validate - 
1. If the SA exist or not
2. If a user has access to SA.

Now, the onus is on ECP or other systems to very carefully set access for every SA?

joel-bluedata:
> Now, the onus is on ECP or other systems to very carefully set access for every SA?

I think that would be true yeah. For example in ECP if there's an SA that you want to be usable by tenant admins/members, then a role binding for those groups would need to give permissions on all SAs in the namespace, or on that particular SA. I think that sound like the correct arrangement in any case yeah?

> Can we somehow use SubjectAccessReview for this to make sure the user is authorized to access the specified serviceAccount. userInfo in admissionRequest should have the information needed?

Well let's have a quick look.

We can use ar.Request.UserInfo to get the name/uid/groups of the user making the KD request, in a validation handler like admitClusterCR. The various fields of UserInfo are marked as "optional" ... I'm not sure under what situations we could _not_ depend on them being present, but yeah if they're normally there we could do a SubjectAccessReview.

The process for that would be:

* Compose a SAR object with the user attributes, and the necessary resource attributes for the SA in question.

* POST the SAR using our shared.Create function.

* After doing the Create, the SAR object that we sent as an argument should now have its Status filled in I think. Some interesting quirks in that status object but I guess we would only reject the request if "denied" is true.

That could be done during kdcluster admission for kdcluster create or edit.

Seems plausible and not too much churn.

(If this works cleanly we should think about doing it for other kdcluster mounts or connections as well, in a later PR.)

kmathur2:
> > I dunno... @swamibluedata did you have something in mind?
> 
> Can we somehow use SubjectAccessReview for this to make sure the user is authorized to access the specified serviceAccount. userInfo in admissionRequest should have the information needed?

Done

joel-bluedata:
Yup looks good IMO. We can chat about it on the call today if there's anything else to say.

joel-bluedata (approved):

Issue #512, opened 2021-09-24
better handling of masterEncryptionKey change


joel-bluedata:
Currently if the masterEncryptionKey in kd-global-config is changed, the following things are true:

- Any newly specified secret key value will be properly encrypted.
- Any existing value will still be readable within the container as long as nothing triggers a configmeta update.
- Any existing value will get scrambled within the container when configmeta update happens. (Unless a change in master key length actually now causes any configmeta update to error out, which is even worse.)

The first of those is fine and good. The second is probably good/expected too. The third is not. And the third isn't easy to catch/fix in the validator because we no longer have either the original value or the original master key accessible to KD (at least, not easily).

Another consideration here: why would someone WANT to change the master key? The only reason that immediately presents itself is that someone-who-shouldn't got a look at the master key and therefore they would be able to decrypt the encryptedKey values in the existing kdcluster specs. Well, even if you change the master key, those encryptedKey values will not automatically be changed, so they would still be vulnerable to anyone who stole the old master key.

Let's discuss the desired behavior, and if there's any low-hanging-fruit improvement we need to make before the 0.7.0 release.

My thought about immediate changes:

Failure to decrypt an encryptedValue should just remove the associated value from the configmeta, rather than failing the configmeta generation.

Possibly we should block changes to masterEncryptionKey while kdclusters exist. (This would be quick to do, just check if shared.clustersUsingApp has any content.) This is super annoying since the only valid reason to check masterEncryptionKey occurs when some kdclusters exist ... however as mentioned above, we don't yet properly handle that situation anyway.

joel-bluedata:
Examples of those "immediate changes" added to PR #513.

joel-bluedata:
Merged a few immediate changes/fixes:
* No changing the master key while kdclusters exist.
* No deleting kdconfig while kdclusters exist (because basically this changes the master key).
* If a secret-key value fails to decrypt, just omit it from configmeta rather than failing all configmeta generation.

Now we need to decide how we can support changing the master key.

PR #513, opened 2021-09-24
secretKeys documentation and fixes

joel-bluedata:
The main point of this PR is to add a doc about "secret keys" that can be referenced from the CRD wiki pages.

I've also bundled a fix for an upgrade case: make sure that the master key default is generated if there was a pre-existing kdconfig when KD was upgraded.

I think it's quite likely that we'll want to include some immediate fixes for issue #512 in here as well, along with the necessary doc updates to reflect that.


joel-bluedata:
The last two commits are proposed near-term fixes for the changes described in issue #512, so everyone can see what code changes would be involved. If they're acceptable I'll keep them in here and update the doc accordingly.

doc/secret-keys.md
```diff
@@ -0,0 +1,91 @@
...
+An example of accessing the value for the secret key named "some-key-name", using the configcli Python module:
+```python
+from configcli import ConfigCli
+configcli = ConfigCli(shell=False)
+namespace = configcli.getCommandObject("namespace")
+distroID = namespace.getWithTokens(["node", "distro_id"])
+ngID = namespace.getWithTokens(["node", "nodegroup_id"])
+roleID = namespace.getWithTokens(["node", "role_id"])
+secretKeyName = "some-key-name"
+secretKeyValue = namespace.getWithTokens(["distros", distroID, ngID, "roles", roleID, "secret_keys", secretKeyName])
```
kmathur2, line 78: This will give encrypted value only, right? App devs will have to decrypt?

kmathur2 (approved):

doc/secret-keys.md
kmathur2, line 78: @aleshchynskyi @joel-bluedata  If I remember this correctly, this still gives decrypt value? 

Can we clarify this in the doc.

joel-bluedata: Nope, this will give unencrypted... only the unencrypted value is stored in the configmeta.

Elaborating a bit: this is the value that was originally passed in as the "value" in a secretKeys element. Now it may be the case that this value is ITSELF a decryption key that will be used to decrypt other things (e.g. from a Secret or Configmap), but whatever that usage is, it's application-specific.

Issue #515, opened 2021-09-30
SubjectAccessReview for SA access needs to include "Extra"


joel-bluedata:
In validateRoleServiceAccount, the SubjectAccessReviewSpec copies User, Group, and UID from the UserInfo. It also need to copy Extra.

(Also please double-check that this SubjectAccessReview works correctly if the client is itself using a service account to authenticate. Edit: tested this part myself just now, does look correct when using SA token to authenticate.)

PR #516, opened 2021-10-01
Add Extra values from authenticator to subject access review

kmathur2:
https://github.com/bluek8s/kubedirector/issues/515

### DIFF
--- a/pkg/validator/cluster.go
+++ b/pkg/validator/cluster.go
@@ -539,6 +539,11 @@ func validateRoleServiceAccount(
 				"service account "+role.ServiceAccountName+" requested by role "+role.Name+" does not exist")
 			continue
 		}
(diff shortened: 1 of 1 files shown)


pkg/validator/cluster.go
```diff
@@ -539,6 +539,11 @@ func validateRoleServiceAccount(
 				"service account "+role.ServiceAccountName+" requested by role "+role.Name+" does not exist")
 			continue
 		}
+		// Convert k8s.io/api/authentication/v1".ExtraValue -> k8s.io/api/authorization/v1".ExtraValue
+		var xtra map[string]sar.ExtraValue
```
joel-bluedata, line 543: I think to avoid assignment to a nil map this would need to be:
```golang
xtra := make(map[string]sar.ExtraValue)
```

(Unfortunately I don't know off the top of my head how to actually stimulate the API server into putting values in this field for testing.)

joel-bluedata (approved):

Issue #507, opened 2021-07-12
typo(?) in NVIDIA_VISIBLE_DEVICES environment var


joel-bluedata:
We're setting NVIDIA_VISIBLE_DEVICE (rather than NVIDIA_VISIBLE_DEVICES); cf. pkg/executor/types.go

jear:
My understanding is the following : Kubedirector is working today with any string, nvidia.com/gpu , or A100 MIG canonical naming, for instance nvidia.com/mig-3g-20gb. A test would be to modify  NVIDIA_VISIBLE_DEVICE to NVIDIA_VISIBLE_DEVICES, and check if KD is still able to handle A100 MIG devices. I will upgrade to 0.6.2 to test nodeSelector which is usually used with MIG. I think that managing GPU and MIG tenant quota can be tricky due to a CUDA limitation (P2P = Peer-to-peer) that prevents a pod to have 2 MIGS of the same board and be able to launch 2 independant workload, as we are used to do with 2 usual GPU.

joel-bluedata:
Currently KD will set that environment var to VOID if there are no resources of type "nvidia.com/gpu".

This is not having any effect because of the current typo, but if we fix the typo, then suddenly you'll be seeing NVIDIA_VISIBLE_DEVICES set to VOID even if the resource type is for example "nvidia.com/mig-3g-20gb". Would that be problematic?

We could change the behavior to for example instead check for any resource types with the prefix "nvidia.com/" but I want to make sure that would be correct behavior.

Also someone in email mentioned the CUDA_VISIBLE_DEVICES variable and I want to understand when/if that should be explicitly set by KD as well.

jear:
I think it can work. Regarding CUDA_VISIBLE_DEVICES, my understanding is that this is used by TEnsorflow, unkess I miss something.

jear:
My understanding for MIG quota and nvidia GPU operator in imported cluster, containing both MIG-capable and non-MIG-capable GPU…

As soon as you enable MIG, for all GPU type, we should not rely anymore only  on nvidia.com/gpu string in requests/limits.

We should  add  other labels that the operator has added on the node.


# Identify the node with the A100 by playing with jq
kubectl get node -o json | jq '.items[].metadata.labels' | grep machine
  "nvidia.com/gpu.machine": "ProLiant-BL460c-Gen8",
  "nvidia.com/gpu.machine": "ProLiant-DL380-Gen10",
[...]

PR #517, opened 2021-10-13
#507 Use NVIDIA_VISIBLE_DEVICES=VOID workaround for all nvidia resources

aleshchynskyi:


aleshchynskyi:
> Looks plausible to me. Have you been able to functionally verify that GPU-using apps and MIG cases still work?

I don't have an environment with MIGs, so currently just an ordinary test cases

joel-bluedata:
OK let's come back to this one. From the discussion/updates on issue #507 I think this is pretty much what we want, except maybe setting a value of "void" as opposed to "VOID".

Need to get an environment to test.

joel-bluedata:
closes #507

joel-bluedata (approved):
Merging this to enable easier/wider testing.

PR #538, opened 2022-01-07
Ezml 862 allow role image upgrade

Kosta91:
As I could see, the nessesary changes for live upgrade for KDApp was ignoring differences between ImageRepoTag and DefaultImageRepoTag at the validation stage. With applied changes in app.go there is possible to replace KDApp spec and then all new KDCluster instances will be created with new images.


joel-bluedata:
I've created the live-app-upgrade branch that we can use to collaboratively stash incremental work on this feature. So I'll close this PR and we can open PRs against that branch. Also cf. our Slack discussion for comments about the changes.

PR #539, opened 2022-01-10
Ezml 862 allow role image upgrade

Kosta91:
Implemented live KDApp spec image upgrading.
Added ability for rolling update statefulset, when corresponding role image was changed.

### DIFF
--- a/pkg/catalog/interrogate.go
+++ b/pkg/catalog/interrogate.go
@@ -22,6 +22,7 @@ import (
 	kdv1 "github.com/bluek8s/kubedirector/pkg/apis/kubedirector/v1beta1"
 	"github.com/bluek8s/kubedirector/pkg/observer"
 	"github.com/bluek8s/kubedirector/pkg/shared"
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 )
 
@@ -194,6 +195,13 @@ func ImageForRole(
 		role,
 		cr.Spec.AppID,
 	)
+}
+
+// ImageForStatefulSet returns the image is currently used at a given statefulSet
+func ImageForStatefulSet(
+	statefulSet *appsv1.StatefulSet,
+) string {
+	return statefulSet.Spec.Template.Spec.Containers[0].Image
 }
 
 // AppSetupPackageURL returns the app setup package url for a given role. The
--- a/pkg/catalog/interrogate.go
+++ b/pkg/catalog/interrogate.go
@@ -22,7 +22,6 @@ import (
 	kdv1 "github.com/bluek8s/kubedirector/pkg/apis/kubedirector/v1beta1"
 	"github.com/bluek8s/kubedirector/pkg/observer"
 	"github.com/bluek8s/kubedirector/pkg/shared"
-	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 )
 
@@ -195,13 +194,6 @@ func ImageForRole(
 		role,
 		cr.Spec.AppID,
 	)
-}
-
-// ImageForStatefulSet returns the image is currently used at a given statefulSet
-func ImageForStatefulSet(
-	statefulSet *appsv1.StatefulSet,
-) string {
-	return statefulSet.Spec.Template.Spec.Containers[0].Image
 }
 
 // AppSetupPackageURL returns the app setup package url for a given role. The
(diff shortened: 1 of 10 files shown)


pkg/catalog/interrogate.go
```diff
@@ -196,6 +197,13 @@ func ImageForRole(
 	)
 }
 
+// ImageForStatefulSet returns the image is currently used at a given statefulSet
```
joel-bluedata, line 200: Probably makes more sense to have this in pkg/shared/util.go ... not really a catalog-related thing.

pkg/executor/statefulset.go
```diff
@@ -149,26 +149,52 @@ func UpdateStatefulSetNonReplicas(
...
-	)
-	// So, what to do. Do we add our owner ref to the existing ones? What if
-	// something else is claiming to be controller? Probably some stale ref
-	// left by a bad backup/restore process? We're just going to nuke any
-	// existing owner refs.
-	patchedRes := *statefulSet
-	patchedRes.OwnerReferences = shared.OwnerReferences(cr)
-	patchErr := shared.Patch(
+
+	patchErr := shared.Update(
```
joel-bluedata, line 196: I'm really puzzled why Patch didn't work for you... we'd need to explore this a bit more.

A problem with Update is that it will fail if the resource has been modified since it has been read (either by some other code in KD or by the K8s statefulset controller or by whoever). This is not rare, and it means that the whole handler will fail and the next handler will have to try again. This is annoying at best, and at worst it's possible with some kinds of updates to get hardlocked and never succeed because (for this or that reason) one part of KD always updates the resource before some other part tries.

Anyway... patch is nice! Let's see if we can get it to work.

pkg/validator/app.go
```diff
@@ -463,6 +463,25 @@ func admitAppCR(
...
 			prevAppCR.Spec.DefaultSetupPackage = appCR.Spec.DefaultSetupPackage
+
+			// EZML-862
+			// Before doing the comparison, we should ignore defferences
+			// between the images defined for different roles. The app configuration
+			// should be able to replace if the image is changed
+			// even when some KD cluster instance based on this spec is alive.
+			if appCR.Spec.DefaultImageRepoTag != nil {
+				*prevAppCR.Spec.DefaultImageRepoTag = *appCR.Spec.DefaultImageRepoTag
+			}
```
joel-bluedata, line 474: I'm dubious about whether we should allow setting the default image for an existing kdapp. The semantics of default image are:

At create time, if any role does not specify an image then it will get the default image. The default image is then removed from the spec.

So if we modify default image for an existing kdapp it will never have any effect, since all roles will have an image by that time. That's probably counter to what folks expect.

pkg/executor/statefulset.go
joel-bluedata: OK so yeah, I think it's a shallow-copy problem. On the line above we just do "patchedRes := *statefulSet" which only copies over the top-level properties of the statefulset in order to make the new object. This was fine when we were just updating the owner ref, but now that we're digging down into the slice of containers it's not fine. When we modify the containers slice, the modification is seen by both the old and new object.

So before doing that modification, we need to deep-copy the container slice and assign that copy to the patchedRes. One example of doing that sort of thing is for the annotations slice modified by SetBackupAnnotation in pkg/executor/cluster.go.

pkg/catalog/interrogate.go
Kosta91: Well, I removed it and added to `pkg/shared/util.go` more common function `StatefulSetContainers` returns the containers array of a statefulset spec

pkg/executor/statefulset.go
Kosta91: Got it, I changed the container-related logic and it looks like that `Patch` works

pkg/validator/app.go
Kosta91: Removed

pkg/shared/util.go
```diff
@@ -111,3 +113,10 @@ func OwnerReferencesPresent(
 	}
 	return false
 }
+
+// StatefulSetContainers returns the array of containers are run for a given statefulSet
+func StatefulSetContainers(
+	statefulSet *appsv1.StatefulSet,
+) []v1.Container {
+	return statefulSet.Spec.Template.Spec.Containers
```
joel-bluedata, line 121: Just to match the existing code format, let's have one blank line at the start of the function body.

joel-bluedata:
Looks good, honestly I think the only necessary change is that blank line. :-) Seems weird to stall the merge for that but we're not in rush, so I'll wait for that change and then merge.

pkg/shared/util.go
Kosta91: Done

PR #541, opened 2022-01-13
Ezml 865 upgrade notification

Kosta91:


deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
```diff
@@ -136,7 +136,7 @@ spec:
                     type: array
                     items:
                       type: string
-                      pattern: '^configure$|^addnodes$|^delnodes$'
+                      pattern: '^configure$|^addnodes$|^delnodes$|^upgrade$'
```
joel-bluedata, line 139: Not related to your change, but I see a bug here where this list should also include the reconnect event. Let's not fix it here though, I will file a separate JIRA to not only fix the spec but also examine the reconnect event code to make sure it honors the eventList spec.

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
```diff
@@ -65,6 +65,7 @@ type KubeDirectorClusterStatus struct {
 	LastNodeID              int64            `json:"lastNodeID"`
 	Roles                   []RoleStatus     `json:"roles"`
 	LastConnectionHash      string           `json:"lastConnectionHash"`
+	UpgradedRoles           map[string]bool  `json:"upgradedroles,omitempty"`
```
joel-bluedata, line 68: I need to read/comment more about how this info is used, but a few top level comments:

* Rather than sticking the info up here, I would recommend to put it as a new status field (or just a boolean flag perhaps) in the RoleStatus object, so that each role's status reflects whether it is currently being upgraded. It makes it a little more faffy for our code to write and read the correct RoleStatus object based on the role name, but we have utility functions to take care of that.

* Regardless of where we add the status info, we do also need to correspondingly update the status stanza in the kdcluster CRD.

* Shouldn't we clear/reset this info once the role upgrade has completed? Or more broadly speaking, how is this handled for multiple upgrades in sequence?

* If this info is meant to reflect whether a role is currently being upgraded (and hasn't finished yet), then its name will probably need to be different, e.g. using the term "upgrading" rather than "upgraded".

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -118,6 +118,9 @@ func syncMembers(
 		if _, ok := r.membersByState[memberCreating]; ok {
 			handleCreatingMembers(reqLogger, cr, r, roles, configmetaGenerator)
 		}
+		if _, ok := r.membersByState[memberUpgrading]; ok {
+			handleUpgradingMembers(reqLogger, cr, r, roles)
+		}
```
joel-bluedata, line 123: A general thought here...

Modifying the member state machine is potentially pretty tricky to get correct in all cases. It may be something we do need to do in this case, but maybe there's a different way.

When a pod gets upgraded, its container restarts, so KD kicks it back into the create pending state, and then it eventually gets to creating state.

When KD handles a creating-state member, if it has an app setup package then it calls appConfig.

appConfig checks to see if the initial config needs to be run (or re-run, in the case of a previous error). If it does need to run initial config then it runs the startscript with the "--configure" arg (the appPrepConfigRunCmd).

(I think in this case, if for whatever reason we're running initial config, we should not send an upgrade notify to this pod.)

Let's say though that appConfig determines that it does NOT need to run the initial config, because it sees that has alreayd been run on this pod. In that case it returns without doing anything. What it seems like we could do, is modify this part of the process. For example:

If appConfig sees that config has already been run previously, it can check whether the container image has changed (e.g. by referencing that new status data you're adding). If so, then instead of doing nothing, it can run the startscript with the "--upgraded" argument.

Something to think about. If it's possible to avoid adding new member states (with roughly the same or nicer code complexity) then IMO that is preferable.

pkg/validator/app.go
```diff
@@ -306,21 +306,42 @@ func validateRoles(
...
+		// If the role spec doesn't contain individual persistent dirs
+		// copy them from global application spec
+		if role.PersistDirs == nil && globalPersistDirs != nil {
+			role.PersistDirs = globalPersistDirs
+		}
+
+		// EZML-865
+		// For each persistent dir defined for the role add
+		// corresponding old_<directory_name> persistent dir for backup data
+		// for upgrade
```
joel-bluedata, line 319: We can talk about this in the meeting, curious to hear more about the thoughts behind this.

joel-bluedata: As discussed in meeting, let's omit this aspect for now.

Kosta91: Discarded

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
Kosta91: Discarded this change

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
Kosta91: @joel-bluedata , I decided to try add `UpgradingMembers` map field to `RoleStatus` instead of `UpgradeRoles` map. It should be filled on the `syncRoles()` stage as `memberName -> <new role image tag>` key-value pairs, before patching the statefulset. Then, on the `syncMembers()` stage, it will be sequentially cleaned, pod by pod, when we will succesfully notify the current pod container to upgrade. 

Sure, I could made it just as `[]string`, but I prefer to work with maps than with string arrays.
In other option, it could be changed to the boolean `Upgrading` field of `MemberStatus` struct. `RoleStatus` struct contains the `Members []MembersStatus` field, so it is possible to iterate this array twice - at enabling `Upgrading` of each member to true and then at disabling to false, when upgrading notification succeded.

pkg/controller/kubedirectorcluster/members.go
Kosta91: Well, I removed the `memberUpgrading` state and added sending upgrade notification to `appConfig`, as you recommended. Hope, I chose the correct place for it.

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
```diff
@@ -194,10 +194,11 @@ type BlockStorage struct {
...
 type RoleStatus struct {
-	Name                string            `json:"id"`
-	StatefulSet         string            `json:"statefulSet"`
-	Members             []MemberStatus    `json:"members"`
-	EncryptedSecretKeys map[string]string `json:"encryptedSecretKeys,omitempty"`
+	Name                string             `json:"id"`
+	StatefulSet         string             `json:"statefulSet"`
+	Members             []MemberStatus     `json:"members"`
+	EncryptedSecretKeys map[string]string  `json:"encryptedSecretKeys,omitempty"`
+	UpgradingMembers    map[string]*string `json:"upgradingMembers,omitempty"`
```
joel-bluedata, line 201: We also need to reflect this change in kubedirector.hpe.com_kubedirectorapps_crd.yaml

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -444,7 +444,6 @@ func handleCreatingMembers(
 	allRoles []*roleInfo,
 	configmetaGenerator func(string) string,
 ) {
-
```
joel-bluedata, line 447: Let's put this blank line back (for code format consistency).

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -1283,11 +1312,13 @@ func queueNotify(
 	// FQDNs of the affected members.
 	op := ""
 	deltaFqdns := ""
+
+	// At the time this function is called, members in this list are
+	// marked as creating, ready, upgrading or config error. The fqdnsList function
+	// will appropriately skip the ones that are still creating, or the
+	// ones in other states that are just reboots.
+
```
joel-bluedata, line 1320: This comment should be put back where it previously was; it refers to the modifiedRole.membersByState[memberCreating] list.

pkg/executor/statefulset.go
```diff
@@ -149,23 +149,71 @@ func UpdateStatefulSetNonReplicas(
...
+		// Set current role members count should be upgraded
+		// It will be used at the syncMembers() step
+		rs, err := shared.GetRoleStatusByName(cr, role.Name)
+		if err != nil {
+			return err
+		}
+
+		if (*rs).UpgradingMembers == nil {
+			(*rs).UpgradingMembers = make(map[string]*string)
+		}
```
joel-bluedata, line 204: Thinking about the implications of the situation where UpgradingMembers already has some elements populated here. That would be bad, as I don't think we should be upgrading a role while a previous upgrade to that role is still cranking along.

In the current state of the live-app-upgrade branch, you have it so that this image reconciliation stuff will be skipped if the cluster is not in a ready state. That's good! But the cluster's overall state doesn't reflect whether some individual role is upgrading, so it won't protect us here.

We need to think/talk about the correct approach. Some possibilities:

- Skip doing the image reconciliation for this role if UpgradingMembers is non-empty.
- In the validator, reject changing the role image if that role's UpgradingMembers is non-empty.
- ??? probably other things we could do also.

At first glance I'm leaning toward that validator check as the correct solution.

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -1158,6 +1154,39 @@ func appConfig(
 				}
 				status, convErr := strconv.Atoi(configStatus)
 				if convErr == nil && status == 0 {
+					rs, err := shared.GetRoleStatusByName(cr, roleName)
+					if err != nil {
+						return true, err
+					}
+					upgradingMembers := (*rs).UpgradingMembers
+
+					if upgradingMembers != nil && upgradingMembers[podName] != nil {
```
joel-bluedata, line 1163: So, here you check for a nil value to decide whether to notify, and also below you set the value to nil when notify has been sent.

Rather than using nil value for this it might be better to use "absence of value" ... so here you would check whether the upgradingMembers[podName] bucket exists in the map, and below you would delete it out of the map. A couple of reasons for that:

- When someone (person or code) is looking at this map to see if a member is currently being upgraded, it's more natural IMO to just see if the member is in the map rather than also having to check the value.
- However, more importantly, I think we can already get into a situation where there are members that aren't in this map, i.e. the role is expanded after the upgrade has been started. So a new member is added to the role (using the new image) and configured, but it is not in this map since it didn't exist when the upgrade was requested. In the current state of the code, that will cause an error on this line when reading upgradingMembers[podName].

So yeah, mostly because of that second reason it seems better to use absence-of-bucket as the representation of "not being upgraded".

joel-bluedata: I'm also thinking about whether there is any case where we can have members left over in this map, that don't get cleared out. That would only be the case for members that are in the process of being set up when the upgrade is triggered. However, that shouldn't be something that can happen since the role upgrade can only start while the cluster is in a stable ready state.

I think that's all good but I will ponder it a bit more.

joel-bluedata:
Looking good! I think this is pretty close, a few comments above.

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -1006,9 +1005,6 @@ func generateNotifies(
 				if member.StateDetail.LastSetupGeneration == nil {
 					continue
 				}
-				if *member.StateDetail.LastSetupGeneration == *cr.Status.SpecGenerationToProcess {
-					continue
-				}
```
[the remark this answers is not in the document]
Kosta91: My bad, guess it happaned by some IDE trouble. Reverted.

pkg/controller/kubedirectorcluster/members.go
Kosta91: Fixed

pkg/controller/kubedirectorcluster/members.go
Kosta91: Done

pkg/controller/kubedirectorcluster/members.go
Kosta91: Done

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
Kosta91: Pardon, `kubedirector.hpe.com_kubedirectorapps_crd.yaml` or `kubedirector.hpe.com_kubedirectorclusters_crd.yaml` ?

pkg/executor/statefulset.go
Kosta91: So, should we reject a role changing for all clusters because of some member of some cluster is in the upgrading state?

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
Kosta91: Actually I updated the `kubedirectorclusters` file, as RoleStatus struct is a part of KDCluster, not KDApp..

pkg/executor/statefulset.go
joel-bluedata: Hmm.

Yeah that does sound like overkill. I guess the validator is not the best approach.

Ignoring the new image until the old image upgrade is finished, that may be the way to go unless some other solution comes to mind. Basically, we could say that UpgradingMembers will be populated here with the entries for this appRoleImage only if UpgradingMembers is currently empty.

Kosta91: Well, I rejected image reconciliation on the statefulset-patching stage. But there appeared the problem, when user may edit wrong image tag or the new image may be unavailable by some reason. It causes the cluster falling into non-ready state + the target role is sealed that it is in upgrading status. The failed pod will endless restart and fail (ErrImagePull), so it may be healed by upgrading or rolling back app spec image again. There are few option how it could be resolved:

- Memorize last working version  and force rolling back the image to it.
- Add ability to change app spec and patch the statefulset again, despite the cluster is not ready.

Currently I chose the second case, as it is IMO more common and user may decide, if he wants just discard appspec changes or change wrong image to some another one. For it I added specific logic, contains:
- Added to `MemberStatus` struct the new string field `ContainerImage` that stores the last succesfully performed image for this concrete pod.
- Before patching a statefulset check, if its role status is in upgrading state and cluster is not ready in the same time. If so, check, if all role member (pods) are marked for upgrade (`UpgradingMembers` map) by the same image as defined in the app spec for this role. If so, do nothing, pod just could updates slowly. Otherwise, (appSpec image is different) it means user changed the role again. In this case we need to full clean the role `UpgradingMembers` map and mark that we need to restart the upgrade.
- Allow statefulset patching by `clusterIsReady && !roleIsUpgrading` flags combination or, alternatively, by `restartUpgrade` flag.
- Note, that if user decide to rollback to the last succesfully run image, it could be still run in the few members of the current role, except the failed one. In this case, we shouldn't add them to role upgrading members, as in this pods will not restart but the role will be always marked as upgrading and there will be not possible to upgrade it in the future. For this purpose we need to compare the `appSpecImage` with `MemberStatus.ContainerImage` field.

joel-bluedata:
I'm generally on board with your changes but I want to spend a bit of time today and tomorrow actually running the code a bit to help me think about any edge cases.

PR #544, opened 2022-01-21
Added pause until validator service start

Kosta91:
Quick-fix for the Makefile deploy target problem, when application examples are started to deploy but `kubectl-validator` service is not ready, which causes the error:
`Error from server (InternalError): error when creating "deploy/example_catalog/cr-app-cassandra-3.11.json": Internal error occurred: failed calling webhook "hard-validate-cr.kubedirector.hpe.com": Post "https://kubedirector-validator.hpecp.svc:443/validate?timeout=30s": service "kubedirector-validator" not found`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,6 +159,9 @@ deploy:
         done
 	@echo
 	@echo
+	while ! kubectl get service kubedirector-validator &> /dev/null; do \
+	    sleep 3; \
+	done
 	@echo \* Creating example application types...
 	kubectl create -f deploy/example_catalog/
 	@echo


Makefile
```diff
@@ -159,6 +159,9 @@ deploy:
         done
 	@echo
 	@echo
+	while ! kubectl get services | grep kubedirector-validator; do \
+	    sleep 3; \
+	done
```
joel-bluedata, line 164: Yeah this is a really tight window you're hitting, if the webhook configuration exists but the service is not created yet. However there is in fact a window there and I guess you're hitting it!

The only tweak I would recommend is just to use the same form of "does this thing exist" check as the other checks in the Makefile. E.g.:
```
	while ! kubectl get service kubedirector-validator &> /dev/null; do \
	    sleep 3; \
	done
```

I guess we could get fancier with a bailout after some time, but really if the webhook configuration exists then the service SHOULD be along really soon.

joel-bluedata (approved):

Issue #537, opened 2022-01-06
prune down default persist dirs


joel-bluedata:
Currently we always persist /etc, /opt, and /usr. On some app images these can be very large, including a bunch of stuff we don't actually care to persist.

Our defaults should be pruned down to just the subdirectories needed by configcli and any generic startscript setup/activities. If an app needs more they can and should specify.

In theory this change should not affect existing apps since they should always specify which directories they want to persist, regardless of whether those directories are among the defaults. In practice though we will certainly break some existing apps if we change the defaults. So we need a way to handle this transition, probably by letting an app explicitly say that it doesn't need the broader defaults.

PR #546, opened 2022-01-25
issue #537: prune down the default persist dirs

joel-bluedata:
This change allows a kdapp resource to opt-in to a different filesystem layout which can make PV initialization much faster.

The best functional description is in the new app-filesystem-layout.md doc contained in this changelist.

The Makefile change here is tangentially related; since a new configcli package is used, I want the build process to make sure that it will get the new configcli if the currently downloaded package is not the right one.

I'll add some other notes as code comments in the changes.

====================

A quick example of the benefits of the new layout. One my usual test system (granted, VMs with slow I/O) launching jupyter notebook with a PV can take hours with the old layout. With the new layout, less than a minute.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -5,9 +5,9 @@ ENV OPERATOR=/home/kubedirector/kubedirector \
     USER_NAME=kubedirector \
     USER_HOME=/home/kubedirector
 
-# Support ps and killall, for redeploy.
+# Support ps, killall, and tar for redeploy.
 RUN microdnf update -y && rm -rf /var/cache/yum
-RUN microdnf -y install --nodocs psmisc procps-ng \
+RUN microdnf -y install --nodocs psmisc procps-ng tar \
     && microdnf clean all
 
 COPY build/bin/entrypoint /usr/local/bin/entrypoint
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
@@ -61,6 +61,8 @@ spec:
                 packageURL:
                   type: string
                   pattern: '^(file|https?)://.+\.tgz$'
+                useNewSetupLayout:
+                  type: boolean
             services:
               type: array
               items:
@@ -127,6 +129,8 @@ spec:
                       packageURL:
                         type: string
                         pattern: '^(file|https?)://.+\.tgz$'
+                      useNewSetupLayout:
+                        type: boolean
                   persistDirs:
                     type: array
                     items:
--- a/pkg/apis/kubedirector/v1beta1/decode.go
+++ b/pkg/apis/kubedirector/v1beta1/decode.go
@@ -36,7 +36,7 @@ func (setupPackage *SetupPackage) UnmarshalJSON(
 		return nil
 	}
 
-	if err := json.Unmarshal(data, &setupPackage.PackageURL); err != nil {
+	if err := json.Unmarshal(data, &setupPackage.Info); err != nil {
 		return err
 	}
 	setupPackage.IsNull = false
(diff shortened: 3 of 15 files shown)


build/Dockerfile
```diff
@@ -5,9 +5,9 @@ ENV OPERATOR=/home/kubedirector/kubedirector \
     USER_NAME=kubedirector \
     USER_HOME=/home/kubedirector
 
-# Support ps and killall, for redeploy.
+# Support ps, killall, and tar for redeploy.
 RUN microdnf update -y && rm -rf /var/cache/yum
-RUN microdnf -y install --nodocs psmisc procps-ng \
+RUN microdnf -y install --nodocs psmisc procps-ng tar \
```
joel-bluedata, line 10: This is actually just something I didn't notice about the recent change to using UBI 8. Apparently tar is no longer included by default, and we need it to support "kubectl cp".

pkg/apis/kubedirector/v1beta1/decode.go
```diff
@@ -36,7 +36,7 @@ func (setupPackage *SetupPackage) UnmarshalJSON(
 		return nil
 	}
 
-	if err := json.Unmarshal(data, &setupPackage.PackageURL); err != nil {
+	if err := json.Unmarshal(data, &setupPackage.Info); err != nil {
```
joel-bluedata, line 39: There are a bunch of mechanical changes in the code to reflect that the configpackage object now contains TWO things (URL string and boolean flag) rather than just the URL. Pretty much any change you see that is substituting "Info" for "PackageURL" is because of that.

swamibluedata: 👍

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
```diff
@@ -127,6 +129,8 @@ spec:
                       packageURL:
                         type: string
                         pattern: '^(file|https?)://.+\.tgz$'
+                      useNewSetupLayout:
+                        type: boolean
```
swamibluedata, line 133: Should we also update our example CRs to set this field

pkg/executor/statefulset.go
```diff
@@ -38,10 +38,28 @@ import (
...
-// that will always be placed on shared persistent storage, if app config is provided
-// for a role
-var appConfigDefaultMountFolders = []string{"/etc", "/opt", "/usr"}
+// appConfigDefaultMountFolders identifies set of member filesystems
+// directories that will always be placed on shared persistent storage, if app
+// config is provided for a role, and if that role's package has
+// useNewSetupLayout set true. Note that the code in getStatefulset assumes
+// this is a superset of the defaultMountFolders list.
+var appConfigDefaultMountFolders = []string{
+	"/etc",
```
swamibluedata, line 47: why not /etc/guestconfig

swamibluedata (approved):

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
joel-bluedata: Yep good idea.

pkg/executor/statefulset.go
joel-bluedata: So, the underlying concern is that any service config changes that the app setup does need to be persisted.

Perhaps ideally this would be part of the app's persistdirs. But no app currently specifies this (because they are -- not unreasonably in this case -- depending on the fact that KD captures it) and if we dropped it from the KD defaults then almost every app would need to specify it. I dunno. I decided just to leave it.

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
joel-bluedata: The annoying thing of course is that any of them may need to get their persistDirs corrected to work in the new layout. I'll see which ones are obvious easy changes first.

swamibluedata: Oh yeah. 
We can just pick one, centos?

joel-bluedata:
Last thing to do before merging is, the jupyter-notebook devs are independently trying it out and will give me feedback.

Issue #547, opened 2022-01-27
make configscript error info available without pod container login


joel-bluedata:
Currently, finding out why a member landed in error status means ssh-ing or exec-ing into the container and examining the stdout/stderr/status files in /opt/guestconfig. These are kind of tedious extra steps, and if the user is an app enduser rather than a developer they might not even have the privileges to do this.

It would be better to expose information through the K8s API when a role member lands in config-error status.

A strawman idea: when putting a role member into config-error, KD also reads the stderr file out of the container and puts it into the member status in some string-valued attribute there. A sort of "additional error info" property. We could also advertise the script's exit status there in an int-valued property.

One concern with that idea is that the stderr file could be large -- in fact there's motivation for an app developer to make it verbose. String properties in our kdcluster status though need to be small, since we have motivation to be concise in our status because of the overall limit on resource document size. These are opposing motivations. We could cap the size of the string we put in the member status, but that might make it nonsensical if we have to truncate a long piece of stderr output.

Other thoughts:

* Maybe the configscript should explicitly write some OTHER file that is meant for having its contents advertised in the member status?

* Or, maybe there's a good way for a K8s API user to request a dump of the configscript's stdout/stderr/status for a particular member? How would that request be submitted? Where would the "dump" go?

So, another strawman idea, taking up that last bullet point above. Let's say that there is a per-role boolean attribute in the kdcluster spec, which can be set/changed at any time. If this flag is true it indicates that the configscript stdout/stderr files and status value should be mirrored to a configmap, with one configmap per affected member (configmap can have the same name as the member pod, for example).

Anyway there should be at least a couple of reasonably useful solutions in this space that can give app developers and users easier visibility into why an app setup script failed.

PR #554, opened 2022-02-09
Added app stderr/stdout reflections to MemberStateDetail struct

Kosta91:
https://github.com/bluek8s/kubedirector/issues/547


deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
```diff
@@ -583,6 +583,10 @@ spec:
                               type: string
                             lastConnectionVersion:
                               type: integer
+                            startScriptStdoutMessage:
+                              type: string
```
joel-bluedata, line 587: We can also declare maxLength for both these properties in the schema here, to advertise that to the client as well as to prevent us from violating that length ourselves (if some bug happens). Note that this will absolutely prevent us from going past that length, which could affect our truncation strategy... I'll have more comments about that elsewhere I think.

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -239,6 +240,48 @@ func syncMemberNotifies(
 	wgReady.Wait()
 }
 
+// EZML-960
```
joel-bluedata, line 243: Note that EZML number (or similar tracking info) doesn't have any meaning in the open-source KD repo here.

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -239,6 +240,48 @@ func syncMemberNotifies(
...
+// extractStartScriptLog extracts the first ~300 symbols
+// from a passed file.
```
joel-bluedata, line 245: I kinda think we should show the tail of the stdout/stderr rather than the head. If the stdout/stderr does exceed the limit we've imposed for the member status here, any important stuff is likely to be just before those streams terminate.

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -239,6 +240,48 @@ func syncMemberNotifies(
...
+
+	maxLen := 300
+	var strB strings.Builder
+	fileExists, fileError := readFileFn(filePath, &strB)
+	if fileError != nil {
+		return nil
+	}
+	var msg string
+	if fileExists {
+		msg = shared.TruncateText(strB.String(), maxLen)
```
joel-bluedata, line 261: Of course, attempts to do punctuation-aware truncation will get slightly weirder when truncating to the tail rather than the head. I'm sure you could make it work, or it's also fine to be less fancy and just do a simple/hard truncation.

As mentioned above we do (I think) want to have an absolute limit imposed in the schema which we cannot exceed, regardless of what punctuation is in the message.

pkg/shared/util.go
```diff
@@ -111,3 +114,25 @@ func OwnerReferencesPresent(
 	}
 	return false
 }
+
+// StatefulSetContainers returns the array of containers are run for a given statefulSet
+func StatefulSetContainers(
```
joel-bluedata, line 119: Looks like this function is not used.

pkg/controller/kubedirectorcluster/members.go
Kosta91: What about to print last few lines instead of characters limit?

deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
Kosta91: I added `MaxLogLinesDump` role property (and default property) to specify the quantity of tail lines should be dumped to the `MemberStatusState` struct fields

pkg/controller/kubedirectorcluster/members.go
Kosta91: Replaced by the issue link

pkg/shared/util.go
Kosta91: Sure, removed

pkg/controller/kubedirectorcluster/members.go
Kosta91: Done, now it prints few last lines (the exact quantity for each role could be defined in an app spec)

deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
joel-bluedata: I'm not sure this should be app-definable. We have limited space in the resource document (K8s API accepts up to 3 MB, etcd backend by default can store up to 1 MB) so by allowing the kdapp author to define some field size we're opening the door for them to break certain sizes of kdcluster.

joel-bluedata: So I guess my opinion is that KD should always have some hard character limit for these fields, regardless of anything else. We could support some fancier processing to try to show whole lines and/or allow the kdapp to specify some (lower) limit, but our own hard limit should be an unbreakable backstop.

Kosta91: Well, I changed `MaxLogLinesDump` to `MaxLogSizeDump` field and related logic, that means the user may strongly limitate the characters quantity to dump. Also I added the logic to show whole lines only are included in this character range.

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
```diff
@@ -63,6 +63,8 @@ spec:
                   pattern: '^(file|https?)://.+\.tgz$'
                 useNewSetupLayout:
                   type: boolean
+            defaultMaxLogSizeDump:
+              type: string
```
joel-bluedata, line 67: I think these two new fields should be schema typed as integer.

pkg/apis/kubedirector/v1beta1/kubedirectorapp_types.go
```diff
@@ -105,15 +106,16 @@ type ServiceEndpoint struct {
...
+	ID             string               `json:"id"`
+	Cardinality    string               `json:"cardinality"`
+	ImageRepoTag   *string              `json:"imageRepoTag,omitempty"`
+	SetupPackage   SetupPackage         `json:"configPackage,omitempty"`
+	PersistDirs    *[]string            `json:"persistDirs,omitempty"`
+	EventList      *[]string            `json:"eventList,omitempty"`
+	MinResources   *corev1.ResourceList `json:"minResources,omitempty"`
+	MinStorage     *MinStorage          `json:"minStorage,omitempty"`
+	ContainerSpec  *ContainerSpec       `json:"containerSpec,omitempty"`
+	MaxLogSizeDump string               `json:"maxLogSizeDump,omitempty"`
```
joel-bluedata, line 118: Assuming we change the type of this to int, then here it will need to be *int so that the validator can tell when it is unset.

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -338,19 +378,42 @@ func handleReadyMembers(
...
+
 				cmd := fmt.Sprintf(appPrepConfigReconnectCmd, containerID)
 
 				cmdErr := executor.RunScript(
 					reqLogger,
 					cr,
 					cr.Namespace,
 					m.Pod,
-					m.StateDetail.LastConfiguredContainer,
+					m.StateDetail.ConfiguringContainer,
```
joel-bluedata, line 404: curious about this change

joel-bluedata:
Almost there, a few minor comments/questions above.

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -239,6 +240,45 @@ func syncMemberNotifies(
...
+// setStateDetailLogs sets the extracted results
+// of startscript executions
+// to corresponding MemberStateDetail fields
+func setStateDetailLogs(
+	readFileFn func(string, io.Writer) (bool, error),
+	stateDetail *kdv1.MemberStateDetail,
+	roleMaxLogSize string,
+) {
+
+	maxSize, _ := strconv.Atoi(roleMaxLogSize)
```
joel-bluedata, line 253: I think there's still a couple of issues here although I could be overlooking some things.

First, I still don't see an overall hard backstop to guard against user-specified values breaking the status stanza (by being too large). To recap a couple of my earlier comments:

> I'm not sure this should be app-definable. We have limited space in the resource document (K8s API accepts up to 3 MB, etcd backend by default can store up to 1 MB) so by allowing the kdapp author to define some field size we're opening the door for them to break certain sizes of kdcluster.
> 
> So I guess my opinion is that KD should always have some hard character limit for these fields, regardless of anything else.

Second, it looks like it's possible for roleMaxLogSize to remain undefined. In the current way things are typed, that means empty-string. That would cause an error here on this line.

I think both of these issues could be solved by having a hard limit imposed by KD, for example 256 chars. The hard limit would come into play if roleMaxLogSize is undefined or if it is larger than the hard limit.

I'm still not entirely convinced of the utility of letting the kdapp developer specify some limit here, but in any case we do need the hard limit as a backstop.

deploy/kubedirector/kubedirector.hpe.com_kubedirectorapps_crd.yaml
joel-bluedata: Hmm. Well for ASCII the distinction is moot, but yeah semantically we're talking about number of characters. So yes integer type would be a good choice, especially since we're dealing with small-ish values. (I don't know why I used 5M as an example above.)

joel-bluedata:
OK we've had a few conversations going on above, so let me summarize what I see as the current (minor) issues here.

The max-log-size properties really need to be typed as integer rather than string. As things currently stand, if a user enters something invalid like "foo" for such a property, KD will accept it and then I imagine it would later incur an exception in setStateDetailLogs when trying to do strconv.Atoi.

If we are requiring something to be an integer we need to validate it as that up-front, and typing it as integer in the schema is the best way to do that.

I see that you had some problems above when trying to spec the fields as integer, with an error message being generated about 
'Invalid value: "string": spec.roles.maxLogSizeDump in body must be of type integer'. I think that was probably because the appPatchValue in validateRoles was using a stringValue; we need to implement and use an integer value field for that patch (as we do with ValueInt in clusterPatchValue) In any case, using integers like this is not an insoluble problem :-) so if there are some weird problems that pop up when typing these fields as integer, show me what the code looks like and we'll fix any such problems.

OK that leads us to the question of what to do with any KD max-log-size default, how that applies when the app specifies its own max log size, and what to do if the app doesn't specify ANY max log size (either top-level default or per-role). After our discussion at the last meeting, my understanding was that we left things like this:

* KD will have a default of 256 chars. This is just a default, not a hard limit.

* If the app does NOT specify defaultMaxLogSizeDump, then the KD default will be used. Note that this defaulting behavior is not something that is schema-specified or handled by the K8s apiserver, rather it is something we have to take care of in our webhook (in validateRoles). So in validateRoles, globalMaxLogSizeDump will always have some value during the per-role processing.

That's it. The KD default limit only comes into play in validateRoles there. No use of it elsewhere, no hard-limiting, etc. Because of this defaulting, the log-dump feature will be available even for "legacy" kdapps that don't specific a max-log-size value.

A kdapp can disable this feature by explicitly specifying a max size of zero.

On the flip side of that, a kdapp can cause problems for itself by specifying a weirdly large value that breaks the kdcluster because the resource size limit is exceeded. If a kdapp does that, that's too bad, I guess it shouldn't do that! We might think of adding another kind of ultimate-sanity-check backstop limit, or putting in some KD smarts so that it omits the log dumps if they cause the status update to fail, but that can be handled in some separate/future PR.

joel-bluedata (approved):

Issue #504, opened 2021-06-24
move to apiextensions.k8s.io/v1 and admissionregistration.k8s.io/v1


joel-bluedata:
We need to start using apiextensions.k8s.io/v1 instead of apiextensions.k8s.io/v1beta1, because v1beta1 will be gone in K8s 1.22.

Need to figure out what the upgrade implications are for existing CRDs.

joel-bluedata:
I don't think there are upgrade implications. We'd have to raise our K8s version support floor to 1.16 though.

joel-bluedata:
Either that, or we could get fancier with our make-deploy and detect supported APIs, using the beta or final CRD API version accordingly.

joel-bluedata:
Similar issue for admissionregistration.k8s.io/v1beta1 objects like MutatingWebhookConfiguration. Not sure when v1 was introduced there, but it's certainly in 1.16 so that would still work as a support floor. That will require (minor) code changes. Fortunately that won't require an upgrade of our operator SDK version since we already use the 1.16 API. We do want to upgrade our operator SDK dependency but it's nice not to have this minor change depend on that.

joel-bluedata:
Working on this at https://github.com/joel-bluedata/kubedirector/tree/apiver

joel-bluedata:
One thing I'm pondering is whether to set x-kubernetes-preserve-unknown-fields in the CRD schemas, to preserve the old behavior where unknown fields are not dropped from incoming CRs. In v1 by default they are dropped and not stored in etcd (and in fact can't even get past kubectl input validation unless you use --validate=false).

There a good reasons to prevent/drop unknown fields, e.g. the discussion in https://kubernetes.io/blog/2019/06/20/crd-structural-schema/ so I probably will leave the new v1 default behavior in place.

PR #563, opened 2022-03-08
deal with some deprecated API

joel-bluedata:
closes issue #504

* Moved to v1 instead of v1beta1 for apiextensions.k8s.io and admissionregistration.k8s.io.
  * We can now run on K8s 1.22 and beyond, but our K8s support floor is raised to 1.16.
  * CRs can no longer have non-schema-approved fields, except in locations where we explicitly approve that behavior (as in resource requirements, to support K8s extended resources).

* Didn't change the AdmissionReview/AdmissionResponse versions quite yet; we don't need to (yet), and there are some issues there to untangle. Did mark that module with a unique import alias "av1beta1" to more easily find and change them in the future.

* Updated docs accordingly.

### DIFF
--- a/deploy/example_catalog/cr-app-deployment-engine.json
+++ b/deploy/example_catalog/cr-app-deployment-engine.json
@@ -15,14 +15,10 @@
         "configSchemaVersion": 7,
         "label": {
             "name": "ML Inferencing",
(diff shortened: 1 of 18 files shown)


joel-bluedata:
BTW may want to delay merging this so we can get a minor release out with just some of the kdcluster status features, and then do a minor release with just this API change. But I need to start getting some eyeballs on it regardless.

swamibluedata (approved):

PR #567, opened 2022-03-17
Ezml 1041 live cluster upgrade

Kosta91:


Kosta91:
During implementation of live application upgrade feature, the next changes were made:

1. KubeDirectorApp types changes:
1.1 `KubeDirectorAppSpec.Upgradable` field.
    type: `boolean`
    json-name: "upgradable"
    `Upgradable` field should be defined by user in the application CRD. It determines, could the application CRD be upgraded
while any live  KubeDirector virtual cluster uses an application version described by this CRD.

2. KubeDirectorCluster types changes:
2.1 `RoleStatus.UpgradingMembers` field
    type: `map[string]*string`
    `UpgradingMembers` map stores the current role members upgrading status.
2.2 `RoleStatus.UpgradeStatus` field
    type: `RoleUpgradeStatus`
    Describes the current upgrade status for the whole role (statefulset). Possible `RoleUpgradeStatus` values could be: 
    empty, `RoleUpgrading`, `RoleRollingBack`, `RoleUpgraded`, `RoleRolledBack`.
2.3 `MemberStatus.UpgradeStatus`
    type: `MemberUpgradeStatus`
    Describes the concrete pod upgrade status. Possible `MemberUpgradeStatus` values could be: 
    empty, `MemberUpgraded`, `MemberRolledBack`.

**Common upgrade/rollback behavior notes**

**1. Cluster upgrade validation.**
    Each KD application CRD contains three fields, which combination ensures a uniqueness of this CRD: 
`Name`, `DistroId` and `Version`.

Each KD cluster bounds with some application over its `appId` field - it corresponds the application name.
A user can edit the specification of already performed KD cluster - even the `appId` field. The `appId` editing triggers a load another
application spec to the KD cluster. But KubeDirector validator provides some limitations for applications, which could be reloaded:

**validator.cluster.validateGeneralClusterChanges** function
1.1 If previous and current `appId` values are the same, skip to step 1.6.
1.2 If `appId` values are different, compare `DistroId` values of current and candidate application spec.
`DistroId` value should stay the same. If no, reject upgrade.
1.3 Compare `Version` values of current and candidate application spec. `Version` values should be different. If no, reject upgrade.
1.4 Check if `Upgradable` field of the current application is set to true. If no, reject upgrade.
1.5 If there no rejections and current cluster has `configured` state (means the current application works fine),
the new `RollbackInfo` object should be created for case, if there will appear some trouble during upgrade process.
The `RollbackInfo` object mirrors the current application spec `Name`, `DistroId` and `Version` fields. The new object is stored in the global hashmap, where key is `cluster.Name+cluster.Namespace`.
1.6 The last validation step is checking `AppCatalog` changes, this behavior part wasn't edited.

**2. Statefulset upgrade and rollback (represented as cluster role)**
    
**statefulset.UpdateStatefulSetNoReplicas** function
    2.1 When the cluster role gets the signal that its members should be upgraded, firstly it checks:
    a) is the cluster currently in `configured` state
    b) is the current role `UpgradingMembers` map is empty (the role is already not in upgrading process)
    If answers are "yes" and "yes", continue to step 2.3. Otherwise, make additional checks, perhaps there is rollback needed.
    2.2 Rollback checks
    2.2.1 Get the `RollbackInfo` object from 1.5 step for the current cluster. If it is nil, do nothing.
    2.2.2 Compare the `RollbackInfo` object fields with the corresponding fields of the current application. If at least
    the one mismatches, do nothing.
    2.2.3 If the currently loaded application spec is strongly the same as described in the `RollbackInfo`, rollback is needed.
    Clean the `UpgradingMembers` map for the current role and set the `rollback` flag to `true`.
    2.2.4 Set the `UpgradeStatus` field of the current `RoleStatus` to `RoleRollingBack`.
    2.3. Role upgrade process may start by two conditions from 2.1 or by `rollback` flag set as `true`.
    2.3.1 Check if a new application image for the current role is different from currently used image. If not, do nothing
    2.3.2 Add to patch for the statefulset the new image from application spec for its containers.
    2.3.3 If it is clean upgrade (not rollback), set the `UpgradeStatus` of the current `RoleStatus` to `RoleUpgrading`.
    2.3.4 If it is clean upgrade, create the new `UpgradingMembers` map for the current `RoleStatus` and fill it by role member names
    as keys and application role image as values.
    2.3.5 Set `needPatch` flag to true and patch the statefulset.
    
**3. Pod notifications.**
    There was extended the application startscript functionality.

**members.handleCreatingMembers** function
    When the each pod was succesfully upgraded or reverted, KubeDirector calls the startscript inside the pod container 
    with `--pod_upgraded` or `--pod_reverted` arguments correspondingly.

When the current pod is succesfully upgraded, a record of the pod name removes from the role `UpgradingMembers` map 
and the pod upgrade status field changes its value to `MemberUpgraded`.
When the last pod of the current role is sucessfully upgraded and there no records left in the role `UpgradingMembers` map,
the role `UpgradeStatus` field changes its value from `RoleUpgrading` to `RoleUpgraded`.
    
When the current pod is succesfully rolled back (reverted),the pod upgrade status field changes its value to  `MemberRolledBack`.
When the last pod of the current role is sucessfully reverted and there no records left to revert, the role `UpgradeStatus` 
field changes its value from `RoleRollingBack` to `RoleRolledBack`.

pkg/shared/util.go
```diff
@@ -121,3 +121,45 @@ func StatefulSetContainers(
...
+}
+
+// ClusterIsReady checks, is cluster state equals to "configured" literal
+func ClusterIsReady(
+	cr *kdv1.KubeDirectorCluster,
+) bool {
+	return cr.Status.State == "configured"
+}
+
+var ClusterRollbackInfoMap = make(map[string]*kdv1.RollbackInfo)
```
joel-bluedata, line 165: Relying on an in-memory piece of state is tricky since we will lose it if KD gets restarted. We either have to store all the state we need in the kdcluster status, or we have to be able to reconstruct it when KD restarts. In this case I think we would need to store any necessary stuff in the kdcluster status.

This is related to the "New kdcluster status fields" part of our shared design/discussion doc, where it's proposed that we will need info at the top level of the status stanza about what the previous kdapp was, if the spec kdapp has been changed and we're in the middle of an upgrade or rollback.

pkg/validator/cluster.go
```diff
@@ -312,13 +312,61 @@ func validateGeneralClusterChanges(
...
+			valErrors = append(valErrors, appModifiedMsg)
+		}
+
+		if prevCrApp.Spec.Version == crApp.Spec.Version {
+			appModifiedMsg := fmt.Sprintf(
+				versionIsNotModified,
+				crApp.Spec.DistroID,
+				crApp.Spec.Version,
+			)
+			valErrors = append(valErrors, appModifiedMsg)
```
joel-bluedata, line 345: Doesn't this trigger if we make some other change to the kdcluster that doesn't touch the app ID?

Also, even if we did change the app ID, is this really an error case or just something we should silently ignore? It would be kind of a weird situation, but it's possible that there are two different app resources with exactly the same distro ID and version. Hmm. Although that really seems like a confusion/problem waiting to happen, and maybe something that our app validator should prevent.

So all that being said, my reaction to this is: this doesn't look like an error situation, it just looks like a situation where we are NOT doing an upgrade or rollback. So we should just skip ahead to any other validation. This is especially true in the case where prevCrApp is the same as crApp. Even if they're different, it's not technically illegal to have two identical kdapp resources, and if we do ever want to prevent that situation from happening then we should enforce that elsewhere.

pkg/validator/cluster.go
```diff
@@ -312,13 +312,61 @@ func validateGeneralClusterChanges(
...
+		}
+
```
joel-bluedata, line 347: Let's say the versions are in fact different though. It seems like we need to be doing a few other checks here:

* If we're in the middle of an upgrade already, make sure we're only rolling back to the previous kdapp (whose identity should be stored in our kdcluster status, as mentioned elsewhere).
* Make sure this request doesn't contain spec changes other than the app change (may need to check this somewhere else).
* If this is an upgrade, do we want to enforce that the new app version is semantically greater than the old?

pkg/validator/cluster.go
```diff
@@ -312,13 +312,61 @@ func validateGeneralClusterChanges(
...
+		// Check, was the previous cluster configured succesfully
+		// If so, write its app identifiers as rollback info for the current one
+		if shared.ClusterIsReady(prevCr) {
+			rbInfo = &kdv1.RollbackInfo{
+				AppID:    prevCr.Spec.AppID,
+				DistroID: prevCrApp.Spec.DistroID,
+				Version:  prevCrApp.Spec.Version,
+			}
+
+			shared.ClusterRollbackInfoMap[rbKey] = rbInfo
```
joel-bluedata, line 366: I mentioned elsewhere that an in-memory map is problematic because we'll lose it if KD restarts. I should also note that it's usually bad to do state changes as part of the validator, since even if our entire validator handler says that a change is OK it might be rejected by some other webhook and never be allowed to reach the K8s API.

So, any recording of state needs to be done in our reconciliation handlers.

For example, let's look at some of the new top level kdcluster status items we were talking about:
* a status field that reflects spec.app... let's just call it status.app here
* a status field that indicates whether an upgrade or rollback (or neither) is active... let's call it status.appChange
* a status field that indicates the app resource that we are/were upgrading from, if upgrade/rollback is active ... let's call it status.prevApp

When the kdcluster is first created, status.app will get populated with a copy of spec.app.

Now someone comes along and edits spec.app to initiate an upgrade. Our validation handler decides it's OK and lets the change go through.

The reconciliation handler sees that spec.app is different from status.app. It also sees that status.appChange is nil, so it knows an upgrade should be started. It can copy status.app to status.prevApp, copy spec.app to status.app, and set status.appChange to "upgrade".

Anyway, you get the gist. The validator just needs to greenlight the change, and then the reconciliation handler updates status and does other necessary actions like changing the statefulset and sending app notifications.

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -1084,10 +1105,42 @@ func appConfig(
...
+			"app config",
+			strings.NewReader(cmd),
+		)
+		if loggingErr && cmdErr != nil {
+			shared.LogErrorf(
+				reqLogger,
+				cmdErr,
+				cr,
+				shared.EventReasonMember,
+				"failed to run startcsript with --{%s} in member{%s} in role{%s}",
```
joel-bluedata, line 1135: typo startcsript

pkg/validator/cluster.go
```diff
@@ -312,13 +312,61 @@ func validateGeneralClusterChanges(
...
+				appNotUpgradable,
+				prevCrApp.Spec.DistroID,
+				prevCrApp.Spec.Version,
+			)
+			valErrors = append(valErrors, appModifiedMsg)
+		}
+
+		// Check, was the previous cluster configured succesfully
+		// If so, write its app identifiers as rollback info for the current one
+		if shared.ClusterIsReady(prevCr) {
```
joel-bluedata, line 359: If the cluster isn't currently in the configured state I believe we should reject the change, as per the "strict" validation scheme from our discussion.

pkg/validator/cluster.go
joel-bluedata: Also, there's other necessary checks although this function may not be the best place to do them.
* A kdcluster spec change that changes the app is not allowed to change anything else (e.g. member counts).
* If an upgrade is in progress, no spec changes are allowed except for rollback.

pkg/executor/statefulset.go
```diff
@@ -171,29 +171,103 @@ func UpdateStatefulSetNonReplicas(
...
+	// Than, the cluster falls into non-ready state and this role status will be always
+	// in upgrading state and reject any try to re-upgrade an image.
+	// For avoiding this situation we should have an ability to re-edit cluster spec
+	// and give a chance to the role status to restart again.
+	// A rollback process will start when a cluster app spec field is equal to the
+	// correcponding RollbackInfo fields, stored for this cluster
+	rollback := false
+
+	// Check, if cluster is not ready and current role status has upgrading members
+	roleStatusIsUpgrading := shared.RoleStatusIsUpgrading(cr, role.Name)
```
joel-bluedata, line 203: I'm not sure how deep I should dig into reviewing this at the moment ... cf. my review of the validator, I think there will need to be changes in the validator that will also ripple into the reconciliation here.

Broadly speaking I'd say that if the validator does its job then this handler can be simpler. We should know by the time we get here:
* the spec is valid
* it's OK to do the upgrade or rollback if the spec/status is indicating such an operation
* the kdapp identified in spec.app gives us the target image regardless of whether we're upgrading or rolling back

pkg/validator/cluster.go
Kosta91: Done. But I left an ability to apply the change for rollback application. A cluster may be in the `updating` state, and the proposed `AppID` should be equal to cluster `Status.UpgradeInfo.PrevApp` field and a `Status.UpgradeInfo.isRollingBack` field should be `false` (as we shouldn't interrupt the performed rollback process)

pkg/validator/cluster.go
Kosta91: Done. Please find the implemented changes in`pkg/controller/kubedirectorcluster/cluster.go`

pkg/validator/cluster.go
Kosta91: Sure, I changed to comparing `crApp != prevCrApp` (received by clusters `AppId` fields)

pkg/controller/kubedirectorcluster/members.go
Kosta91: Fixed

pkg/executor/statefulset.go
Kosta91: Well, I refactored this part

pkg/validator/cluster.go
```diff
@@ -312,13 +312,79 @@ func validateGeneralClusterChanges(
 	valErrors []string,
 ) []string {
 
-	if cr.Spec.AppID != prevCr.Spec.AppID {
-		appModifiedMsg := fmt.Sprintf(
-			modifiedProperty,
-			"app",
+	readyToUpgrade := shared.ClusterIsReady(prevCr)
```
joel-bluedata, line 315: Since this will cover both upgrade and rollback readiness, let's call the variable something else like `okToChangeApp` or something.

pkg/validator/cluster.go
```diff
@@ -312,13 +312,79 @@ func validateGeneralClusterChanges(
...
+	upgradeInfo := prevCr.Status.UpgradeInfo
+	// If changed cluster has actual upgradeInfo (was in Updating state)
+	// we may permit the changes for rollback case only
+	// It means, the current proposed AppId should be the same as the stored in the upgradeInfo object
+	if upgradeInfo != nil && !upgradeInfo.IsRollingBack {
+		readyToUpgrade = cr.Spec.AppID == upgradeInfo.PrevApp
+	}
+
+	if !readyToUpgrade {
```
joel-bluedata, line 324: It seems like this will fire on any change that happens if the cluster is not currently in "configured" state. I.e. we're triggering an error here before we even check to see if the user is trying to change the app. I think this should be moved down inside the `if appModified` block below.

pkg/validator/types.go
```diff
@@ -54,6 +55,11 @@ const (
 	modifiedProperty = "The %s property is read-only."
 	modifiedRole     = "Role(%s) properties other than the members count cannot be modified while role members exist."
 
+	clusterNotReady      = "The %s cluster cannot be upgraded, as currently it is in %s state."
```
joel-bluedata, line 58: We should probably have different messages for the 'cluster can't be upgraded because it is busy with other changes' case vs. the 'cluster is already upgrading and can only be rolled back to the previous app' case.

pkg/validator/cluster.go
```diff
@@ -312,13 +312,79 @@ func validateGeneralClusterChanges(
...
+		// Also, the current app should support upgrade
+		if !prevCrApp.Spec.Upgradable {
+			appModifiedMsg := fmt.Sprintf(
+				appNotUpgradable,
+				prevCrApp.Spec.DistroID,
+				prevCrApp.Spec.Version,
+			)
+			valErrors = append(valErrors, appModifiedMsg)
+		}
+	}
```
joel-bluedata, line 386: It looks like we still need the following check:

While an upgrade is in progress (i.e. was initiated on some previous handler, so upgradeInfo exists), you're not allowed to make any spec changes other than rollback.

pkg/shared/util.go
```diff
@@ -121,3 +121,43 @@ func StatefulSetContainers(
...
+	}
+
+	return len(rs.UpgradingMembers) > 0
+}
+
+// ClusterIsReady checks, is cluster state equals to "configured" literal
+func ClusterIsReady(
+	cr *kdv1.KubeDirectorCluster,
+) bool {
+	return (*cr).Status.State == "configured"
```
joel-bluedata, line 162: Need a blank line here for code convention.

Although, it looks like currently this function is only called from one spot, so maybe we don't need to define this as a function?

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -57,6 +57,65 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
 		cr.Status = &kdv1.KubeDirectorClusterStatus{}
 		cr.Status.Roles = make([]kdv1.RoleStatus, 0)
 	}
+
+	// Check, if some statefulset is in the middle of upgrade/rollback process.
```
joel-bluedata, line 61: All this new logic added to syncCluster should be moved AFTER the big defer func definition below, just in case for some reason we want/need to do an early return out of syncCluster.

Also this function is already (too) big, so new logic should go into a subroutine e.g. something like handleUpgrade.

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -57,6 +57,65 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
...
+	// If so, the current cluster state still can be in `configured` state, change it to `updating`
+	upgradeIsActive := false
+	for _, rs := range cr.Status.Roles {
+		if rs.RoleUpgradeStatus == kdv1.RoleRollingBack || rs.RoleUpgradeStatus == kdv1.RoleUpgrading {
+			upgradeIsActive = true
+			cr.Status.State = string(clusterUpdating)
```
joel-bluedata, line 67: Hmm. I hadn't really thought about changing the top-level cluster state, but it does make sense. I'm curious though why it's necessary to flip it to "updating" here since it should already have been changed to "updating" when the upgradeInfo was created. That may be a sign that something is wrong about the flow through this overall syncClusters function.

Normally at the bottom of syncClusters, if the current cluster state is NOT "configured" then the code looks to see if syncClusterRoles has declared the need to create/delete any members (by the state value it returns). If not, then the logic is OK, our updating must be finished, so let's set the cluster state back to "configured". I think that may be what is happening to you.

(This would all be a little bit easier to see if we factor our the connections-related code there into a subroutine.)

I think we need to continue to restrict our cluster-state-change logic to that code near the end of syncCluster, and just update that logic.

The current code there looks like this:
```go
	// This "state" is from syncClusterRoles.
	if state == clusterMembersStableReady {
		if cr.Status.State != string(clusterReady) {
			// logging and a bunch of connections stuff
			cr.Status.State = string(clusterReady)
		}
		if currentHash == cr.Status.LastConnectionHash {
			// e.g. if we're not updating connections, leave the cluster as
			// "configured" and we're done.
			return nil
		}
	}

	// We reach this spot if syncClusterRoles determines that members are
	// changing, or if the connections changed (detected by that hash change).
	// In that case, unless we're currently in "creating" state we switch to
	// "updating"
	if cr.Status.State != string(clusterCreating) {
		cr.Status.State = string(clusterUpdating)
	}
```

So maybe it should look something like this:
```go
	upgradeOrRollbackActive := (cr.Status.UpgradeInfo != nil)

	if (state == clusterMembersStableReady) && !upgradeOrRollbackActive {
		if cr.Status.State != string(clusterReady) {
			// logging and a bunch of connections stuff
			cr.Status.State = string(clusterReady)
		}
		if currentHash == cr.Status.LastConnectionHash {
			return nil
		}
	}

	if cr.Status.State != string(clusterCreating) {
		cr.Status.State = string(clusterUpdating)
	}
```

If we handle the cluster state management down there -- which should manage both setting it to "updating" for an upgrade/rollback as well as clearing it back to "configured" when done -- then we shouldn't need to handle it anywhere in this chunk logic.

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -57,6 +57,65 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
...
+		}
+		cr.Status.AppID = &cr.Spec.AppID
+		(*cr).Status.State = string(clusterUpdating)
+	} else
+	// If KD cluster upgradeInfo object exists and is not in RollingBack status,
+	// but the current cluster AppID is the same as in the upgradeInfo
+	// it means that rollback is required
+	if upgradeInfo != nil && !upgradeInfo.IsRollingBack && cr.Spec.AppID == upgradeInfo.PrevApp {
+		upgradeInfo.IsRollingBack = true
+		cr.Status.AppID = &upgradeInfo.PrevApp
```
joel-bluedata, line 116: I think we also should swap prevApp back to the app we are rolling-back-from. That will come in useful elsewhere I think. So this code could do:
```go
upgradeInfo.PrevApp = *cr.Status.AppID
cr.Status.AppID = &cr.Spec.AppID
```

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -57,6 +57,65 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
...
+			cr.Status.State = string(clusterUpdating)
+			break
+		}
+	}
+
+	// If all statefulsets completed their upgrade/rollback processes
+	// erase the cluster upgradeInfo object
+	if cr.Status.UpgradeInfo != nil && !upgradeIsActive {
+		// If there wasn't rollback
+		if cr.Spec.AppID != cr.Status.UpgradeInfo.PrevApp {
```
joel-bluedata, line 76: Hmm, app reference management. I think we need to make sure we have a reference on both apps as long as an upgrade or rollback is active. Assuming we are doing the change to prevApp on rollback that I mentioned below, then that means:
* On creating upgradeInfo (starting upgrade), add the reference to the app in cr.Spec.AppID.
* On deleting upgradeInfo (ending upgrade or rollback), remove the reference to the app in upgradeInfo.PrevApp.

pkg/controller/kubedirectorcluster/cluster.go
```diff
@@ -57,6 +57,65 @@ func (r *ReconcileKubeDirectorCluster) syncCluster(
...
+			)
+			// And bind the cluster with the new app
+			shared.EnsureClusterAppReference(
+				cr.Namespace,
+				cr.Name,
+				*cr.Spec.AppCatalog,
+				cr.Spec.AppID,
+			)
+		}
+		cr.Status.UpgradeInfo = nil
```
joel-bluedata, line 92: All roleUpgradeStatus and podUpgradeStatus should also be cleared when the upgrade/rollback is complete, yeah?

pkg/controller/kubedirectorcluster/types.go
```diff
@@ -88,17 +90,51 @@ const (
...
+				cmd = `rm -f /opt/guestconfig/configure.* && ` + cmd
+			}
+			result = &cmd
+			cmdCache[arg] = result
+		}
+		return *result
+	}
+
+	return fmt.Sprintf(cachedConfigCmdTemplate(), containerId)
+}
```
joel-bluedata, line 136: Some good cleanup/factoring!

I don't think the cachedConfigCmdTemplate func needs to be defined though right? Seems like that code could just be executed as part of the GetAppConfigCmd function.

joel-bluedata:
OK!

I have some more comments/questions/ideas but they may all be in the category of things to handle AFTER this PR. I'll post them separately below this comment.

joel-bluedata:
First, I'm a little wary about the status management being spread around so much, although I do recognize that there are places where it's easier to determine that a status change is needed.

So as this change currently stands, it looks like:
* top-level cluster handler manages status.app and status.upgradeInfo
* statefulset reconciliation manages setting per-role status to upgrading/rollingback
* creating-member handler manages setting per-member status to upgrading/rollback
* creating-member handler also manages setting per-role and per-member status to upgraded/rolledback

(And as mentioned elsewhere, nobody finally clears the per-role/per-member status... probably need to do that in the top-level handler when it clears upgradeInfo.)

Hmm.

Well when I write that out it has a certain logic to the structure.

I guess I need think about whether we can push the role status management into syncClusterRoles, since having status management in the executor package is the only part which really itches me. But that's something I can tackle separately.

joel-bluedata:
If you trigger an upgrade, then immediately trigger a rollback, I'm concerned about some possible race conditions in the pod statuses and notifications. handleCreatingMembers will assume PodRollingBack status but maybe it's possible that the pod is coming up at that moment with the image from the upgrade? Rather than rely on kdcluster status here to see what is going on, we might have to actually check the pod image.

joel-bluedata:
I'm still dubious about the need to have upgradingMembers status at the role-status level, and we don't seem to be making use of the image value stored in that map... but I'll suspend further worrying about that until we're sure the above issue is resolved.

joel-bluedata:
And a collection of lesser things -- again, this is all probably for some followup PRs:

* It may be a usability problem that "upgradable" can't be changed for old existing kdapps (if they are in use). Maybe we should allow it to be set if it is missing.

* We should add membersUpgrading and membersRollingBack fields to the member status rollup.

* Should we enforce that the destination app's version string (for an upgrade) must be semantically greater than the old version string? I'm leaning yes.

pkg/validator/cluster.go
Kosta91: Hmm, shouldn't we reject any changes except app editing for rollback while cluster isn't in `configured` state?

pkg/validator/cluster.go
Kosta91: I meant this variable covers any cluster changes readiness, including pods count changing. Firstly it checks, if cluster is `configured`, and then, also checks if a user requested the rollback.

For the rollback case there is an additional check below, inside the `if appModified` block, that the previous and current cluster specs have no differences other than `AppID`

pkg/validator/cluster.go
Kosta91: Isn't this check not enough?

```
	if appModified {
		// Spec change not allowed if a user tries to update spec.app with other spec fields in the same time.
		prevAppId := prevCr.Spec.AppID
		prevCr.Spec.AppID = cr.Spec.AppID

		if !equality.Semantic.DeepEqual(cr.Spec, prevCr.Spec) {
			valErrors = append(valErrors, notOnlyAppModified)
			return valErrors
		}
		prevCr.Spec.AppID = prevAppId
```

pkg/validator/types.go
Kosta91: Done

pkg/shared/util.go
Kosta91: Ok, I removed it

pkg/controller/kubedirectorcluster/cluster.go
Kosta91: Done

pkg/controller/kubedirectorcluster/cluster.go
Kosta91: Don't we need to have the permanently stored info in the cluster status, that some pods / statefulsets were upgraded or rolledBack?

pkg/controller/kubedirectorcluster/types.go
Kosta91: Done

pkg/validator/cluster.go
neogeographica: We discussed this out-of-band just now ... commenting here just to write it down. We do have cases where it is legal to edit the spec of a cluster that is not in configured state, and the enforcement/checking of that legality is handled elsewhere. So we shouldn't accidentally block any such legal cases here.

pkg/validator/cluster.go
neogeographica: Nope because appModified will not be true on subsequent reconciliation handlers during the upgrade.

Kosta91: Well, I adjusted this logic

pkg/controller/kubedirectorcluster/cluster.go
Kosta91: Fixed

pkg/controller/kubedirectorcluster/cluster.go
Kosta91: Done

pkg/validator/cluster.go
joel-bluedata: @neogeographica is me, sorry. Sometimes I'm logged in with my personal account and forget to switch.

That was the discussion we had in our meeting, about cases like reverting a requested members change that can't be fulfilled.

I'll re-review the validator changes from scratch to see if there are remaining issues.

joel-bluedata:
At this point let's recap the validation requirements in our design doc and compare them against changes in validateGeneralClusterChanges in this PR.

First there's the validation for triggering an upgrade or rollback. I'll restate/collapse them together a bit here, throw in some broad categories, and include the parts of the "more strict" validation checks we decided on which apply during this initial trigger.

* If the kdapp in the kdcluster spec is being changed:
  * Check kdcluster state:
    * Is the kdcluster ready (in a configured state)? If not, reject.
  * Check validity of the requested kdapp:
    * Does the new kdapp have the same distroID as the current one? If not, reject.
    * Does the new kdapp have a different version than the current one? The doc says ignore here, but your current reject approach is probably better (can discuss this elsewhere).
    * Does the new kdapp indicate that it supports live upgrade? If not, reject.
    * Does the new kdapp differ from the current kdapp ONLY in its role container images (and name and version)? If not, reject.
      * Could allow cardinality changes in some cases here, as we discussed in the design doc sidebar.
  * Check current upgrade/rollback state:
    * If a rollback is in progress, reject.
    * If an upgrade is in progress, but we are NOT returning to the previous kdapp, reject.
    * If any other spec changes are also being requested along with the kdapp change, reject.

We're omitting the upgrade check for greater semantic version for now.

Then there's the other half of the "more strict" validation checks we decided on:

* If the kdapp is NOT being changed:
  * If an upgrade or rollback is in progress, reject any spec changes.


I think there are some spots where validateGeneralClusterChanges doesn't match up with this.

For example it will reject any spec change other than kdapp if the kdcluster is not in a configured state, which is too strong and rejects some valid changes as we discussed in the meeting.

Also there's no check yet for whether the new kdapp differs from the old one too much (i.e. fields other than container images, like services etc.). That could come in a subsequent PR though.

Other than that I think it matches though? Or at least is very close. Some checks are done in a different order so I might not have judged it all correctly, but the other checks seem to be in there.

We've massaged this PR a lot so it may be about time to merge it, collect the issues, and start some new PRs to address individual things. Would you rather merge it now or take one more pass over validateGeneralClusterChanges?

Also BTW note that the automatic merge check is failing because of "make golint" issues, so those would need to be resolved before merge.

Issue #592, opened 2022-04-19
allow setting spec.upgradable on in-use kdapps


joel-bluedata:
Part of issue #588. Depends on issue #589, which should be closed before these changes here are added.

This is mostly so that existing kdclusters for "old" kdapps can be marked to take advantage of live app upgrade IF it is deemed that the old kdapp doesn't need to respond to any upgrade notifications.

We can change KubeDirectorAppSpec.Upgradable to a pointer type, so we can tell when it is unspecified. (And code that looks at this field should treat unspecified as false.) The kdapp change validator can allow setting this field, even for in-use kdapps, if it is currently unspecified.

PR #598, opened 2022-04-26
Add ability to set Spec.Upgradable field to in-use KD apps

Kosta91:
Resolves https://github.com/bluek8s/kubedirector/issues/592


joel-bluedata:
I think we need to be a little more restrictive here. This change as it stands would always allow changing the value of this field at any time, regardless of its current value.

What would probably be preferable is: only allow setting a value for it if it is currently unset. (This would require changing KubeDirectorAppSpec.Upgradable to a pointer type, to detect the "is unset" situation.)

neogeographica (approved):

joel-bluedata (approved):

Issue #590, opened 2022-04-19
semantic version check for upgrade


joel-bluedata:
Part of issue #588. Depends on issue #589, which should be closed before these changes here are added.

When upgrading from kdapp A to B, we should enforce that the "version" of both kdapps is understandable according to [semantic versioning](https://semver.org/) rules, and that version B is greater than version A.

There are golang packages that should make this pretty simple.

PR #600, opened 2022-04-28
Ezml 1226 versions semantic control

Kosta91:
Resolves https://github.com/bluek8s/kubedirector/issues/590


pkg/validator/cluster.go
```diff
@@ -379,11 +380,14 @@ func validateGeneralClusterChanges(
 		}
 
 		// App version should be different from previous one
-		if prevCrApp.Spec.Version == crApp.Spec.Version {
+		prevVer := semver.New(prevCrApp.Spec.Version)
+		newVer := semver.New(crApp.Spec.Version)
+
```
joel-bluedata, line 385: At this point, prevVer and/or newVer can be nil if the version string is not understood by semver. In those cases we should also reject the upgrade.

joel-bluedata: Actually, my mistake. In those cases, `New` will panic, which is bad for us.

Try using NewVersion (instead of New) as in https://github.com/coreos/go-semver/blob/main/example.go

joel-bluedata:
Also note that Travis CI is reporting that you need to run "make format".

pkg/validator/cluster.go
Kosta91: Changed

joel-bluedata:
Nice.

joel-bluedata (approved):

Issue #591, opened 2022-04-19
on upgrade, check for allowed differences between old and new kdapp


joel-bluedata:
Part of issue #588. Depends on issue #589, which should be closed before these changes here are added.

For the first cut at live app upgrade, the only spec differences we should allow between the old and new kdapp are the role images.

If we're feeling spicy we could also allow differences in role cardinality as long as the current kdcluster member counts are legal under the new kdapp's spec.

PR #602, opened 2022-05-03
Ezml 1127 check app difference

Kosta91:
Resolves https://github.com/bluek8s/kubedirector/issues/591

### DIFF
--- a/pkg/validator/cluster.go
+++ b/pkg/validator/cluster.go
@@ -474,12 +474,12 @@ func validateAppSpecChanges(
 				allowedRoleMembersCount, scaledOut := catalog.GetRoleCardinality(&appRole)
 				prevCrRoleMembersCount := len(prevCrRole.Members)
 
-				validCardinality := prevCrRoleMembersCount == int(allowedRoleMembersCount)
+				isValidCardinality := prevCrRoleMembersCount == int(allowedRoleMembersCount)
 				if scaledOut {
-					validCardinality = prevCrRoleMembersCount >= int(allowedRoleMembersCount)
+					isValidCardinality = prevCrRoleMembersCount >= int(allowedRoleMembersCount)
 				}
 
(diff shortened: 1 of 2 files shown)


pkg/validator/cluster.go
```diff
@@ -434,6 +443,74 @@ func validateGeneralClusterChanges(
...
+// current cluster role members count. Otherwise, user should previously change
+// cluster role members count at a preceded cluster updating.
+func validateAppSpecChanges(
+	prevCr *kdv1.KubeDirectorCluster,
+	prevApp *kdv1.KubeDirectorApp,
+	targetApp *kdv1.KubeDirectorApp,
+) *string {
+
+	for _, appRole := range targetApp.Spec.NodeRoles {
+		// First, check that role image tags match
```
joel-bluedata, line 461: I guess this isn't really a "check" here, but more like making sure we ignore any difference in images.

pkg/validator/cluster.go
```diff
@@ -408,6 +409,14 @@ func validateGeneralClusterChanges(
 				prevCrApp.Spec.Version,
 			)
 			valErrors = append(valErrors, appModifiedMsg)
+			return valErrors
+		}
+
+		// Also, check the new app differs from the current one in the role images and cardinality only
```
joel-bluedata, line 415: Let's change this comment to just say "check that the new app differs only in allowed ways", or something like that, so that if we change the semantics of validateAppSpecChanges in the future we won't have to come back here to fix the comment.

pkg/validator/cluster.go
Kosta91: Changed

pkg/validator/cluster.go
Kosta91: Changed

joel-bluedata (approved):

Issue #595, opened 2022-04-19
audit/cleanup for upgrade status management


joel-bluedata:
Part of issue #588.

The status management is spread out quite a bit, although I do recognize that there are places where it's easier to determine that a status change is needed.

As the branch stands:

* top-level cluster handler manages status.app and status.upgradeInfo
* statefulset reconciliation manages setting per-role status to upgrading/rollingback
* creating-member handler manages setting per-member status to upgrading/rollback
* creating-member handler also manages setting per-role and per-member status to upgraded/rolledback

There's a logic in that structure, but I think we should at least consider whether we can push the role status management into syncClusterRoles, since having status management in the executor package is the only part which really itches me.

I'm also still dubious about the need to have upgradingMembers status at the role-status level, and we don't seem to be making use of the image value stored in that map... but I'll suspend further worrying about that until we're sure that issue #594 is resolved.

Issue #594, opened 2022-04-19
investigate possible upgrade/rollback race


joel-bluedata:
Part of issue #588.

If you trigger an upgrade, then immediately trigger a rollback, I'm concerned about some possible race conditions in the pod statuses and notifications. handleCreatingMembers will assume PodRollingBack status but maybe it's possible that the pod is coming up at that moment with the image from the upgrade? Rather than rely on kdcluster status here to see what is going on, we might have to actually check the pod image.

Issue #593, opened 2022-04-19
revisit roleUpgradeStatus and podUpgradeStatus fields


joel-bluedata:
Part of issue #588.

Currently these status fields are left with upgraded or rolledback values even after the upgrade or rollback completes.

Is this useful information at that point? Could be. I would just like to examine this a bit more.

We can't have an "audit trail" over multiple upgrades/rollbacks in the status stanza (because of finite document size). We could leave this status here as an indicator of what happened in the previous upgrade attempt -- although that may have been a long time ago. If we do leave this status here, perhaps we also need to leave a top-level status describing what that previous upgrade attempt was.

PR #606, opened 2022-05-10
Extracted role status management to controller package

Kosta91:
Resolves https://github.com/bluek8s/kubedirector/issues/595

1. Extracted `RoleStatus` management from `executor` to `controller` package
2. Corrected `upgradeInfo` object elimination. Before removing the object cluster should check, if there was a rollback then all role upgrade statuses should be set to `RolledBack` value or should stay empty, if a corresponding role was not changed by the last upgrade. Possibly it should also resolve https://github.com/bluek8s/kubedirector/issues/594
When upgradeInfo is removed and upgrade process is finished, all the role upgrade statuses are set to `RoleConfigured` (default empty value), which partially resolves https://github.com/bluek8s/kubedirector/issues/593
3. As currently we don't need to use a new image value stored in `UpgradingMembers` map, it was replaced by the simple `MembersUpgrading` counter. At the direct upgrade it is set to the current role members quantity and than decreases when the each role member is upgraded. At the rollback case its value is set to the already upgraded members quantity. When `MembersUpgrading` counter decreases to 0, the corresponding `RoleUpgradeStatus` value changes to `RoleUpgraded/RoleRolledback`.

### DIFF
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
@@ -601,7 +601,7 @@ spec:
                         type: string
                       statefulSet:
                         type: string
-                      membersUpgrading:
+                      upgradingMembersCount:
                         type: integer
                       roleUpgradeStatus:
                         type: string
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
@@ -601,9 +601,8 @@ spec:
                         type: string
                       statefulSet:
                         type: string
-                      upgradingMembers:
-                        type: object
-                        nullable: true
+                      membersUpgrading:
+                        type: integer
                       roleUpgradeStatus:
                         type: string
                       members:
--- a/pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
+++ b/pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
@@ -226,12 +226,12 @@ const (
 
 // RoleStatus describes the component objects of a virtual cluster role.
 type RoleStatus struct {
-	Name                string            `json:"id"`
-	StatefulSet         string            `json:"statefulSet"`
-	Members             []MemberStatus    `json:"members"`
-	EncryptedSecretKeys map[string]string `json:"encryptedSecretKeys,omitempty"`
-	MembersUpgrading    int               `json:"membersUpgrading,omitempty"`
-	RoleUpgradeStatus   RoleUpgradeStatus `json:"roleUpgradeStatus,omitempty"`
+	Name                  string            `json:"id"`
+	StatefulSet           string            `json:"statefulSet"`
+	Members               []MemberStatus    `json:"members"`
+	EncryptedSecretKeys   map[string]string `json:"encryptedSecretKeys,omitempty"`
+	UpgradingMembersCount int32             `json:"upgradingMembersCount,omitempty"`
+	RoleUpgradeStatus     RoleUpgradeStatus `json:"roleUpgradeStatus,omitempty"`
 }
 
 // UpgradeInfo decribes cluster upgrading status
--- a/pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
+++ b/pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
@@ -253,10 +253,8 @@ type RollbackInfo struct {
 type MemberUpgradeStatus string
 
 const (
-	// PodUpgraded means upgrade process is finished
-	PodUpgraded MemberUpgradeStatus = "upgraded"
-	// PodRolledBack means rollback process is finished
-	PodRolledBack MemberUpgradeStatus = "rolledBack"
+	// PodConfigured is default empty value means the pod is in ready state
+	PodConfigured MemberUpgradeStatus = ""
 	// PodUpgrading means the pod is in the middle of upgrade process
 	PodUpgrading MemberUpgradeStatus = "upgrading"
 	// PodRollingBack means the pod is in the middle of rollback process
--- a/pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
+++ b/pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
@@ -224,12 +224,12 @@ const (
 
 // RoleStatus describes the component objects of a virtual cluster role.
 type RoleStatus struct {
-	Name                string             `json:"id"`
-	StatefulSet         string             `json:"statefulSet"`
-	Members             []MemberStatus     `json:"members"`
-	EncryptedSecretKeys map[string]string  `json:"encryptedSecretKeys,omitempty"`
-	UpgradingMembers    map[string]*string `json:"upgradingMembers,omitempty"`
-	RoleUpgradeStatus   RoleUpgradeStatus  `json:"roleUpgradeStatus,omitempty"`
+	Name                string            `json:"id"`
+	StatefulSet         string            `json:"statefulSet"`
+	Members             []MemberStatus    `json:"members"`
+	EncryptedSecretKeys map[string]string `json:"encryptedSecretKeys,omitempty"`
+	MembersUpgrading    int               `json:"membersUpgrading,omitempty"`
+	RoleUpgradeStatus   RoleUpgradeStatus `json:"roleUpgradeStatus,omitempty"`
 }
 
 // UpgradeInfo decribes cluster upgrading status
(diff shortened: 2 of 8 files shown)


joel-bluedata:
I see where you're going here, but we still have code in executor that is calling GetRoleStatusByName and doing things in the rolestatus, which is the key dubiousness we're trying to avoid.

Maybe we can pass in a function that does all of that status work (rather than just this one bit). Or, perhaps preferably, UpdateStatefulSetNonReplicas could additional pieces of information so that the status manipulation can be done after UpdateStatefulSetNonReplicas returns.

For example let's say that UpdateStatefulSetNonReplicas instead of just returning `error` will return `(*string, error)`. Where that new first element would be some value we could call newImageSet with the semantics that it is set to point to appRoleImage if a container image patch was (successfully) done, and nil if no image patch was attempted or the patch failed.

If a non-nil newImageSet is returned from UpdateStatefulSetNonReplicas, then the code in the caller can do all the necessary setting of fields in the role status (UpgradingMembers and RoleUpgradeStatus).

joel-bluedata:
I see where you're going here, but we still have code in executor that is calling GetRoleStatusByName and doing things in the rolestatus, which is the key dubiousness we're trying to avoid.

Maybe we can pass in a function that does all of that status work (rather than just this one bit). Or, perhaps preferably, UpdateStatefulSetNonReplicas could additional pieces of information so that the status manipulation can be done after UpdateStatefulSetNonReplicas returns.

For example let's say that UpdateStatefulSetNonReplicas instead of just returning `error` will return `(*string, error)`. Where that new first element would be some value we could call newImageSet with the semantics that it is set to point to appRoleImage if a container image patch was (successfully) done, and nil if no image patch was attempted or the patch failed.

If a non-nil newImageSet is returned from UpdateStatefulSetNonReplicas, then the code in the caller can do all the necessary setting of fields in the role status (UpgradingMembers and RoleUpgradeStatus).

Kosta91:
> I see where you're going here, but we still have code in executor that is calling GetRoleStatusByName and doing things in the rolestatus, which is the key dubiousness we're trying to avoid.
> 
> Maybe we can pass in a function that does all of that status work (rather than just this one bit). Or, perhaps preferably, UpdateStatefulSetNonReplicas could return additional information so that the status manipulation can be done after UpdateStatefulSetNonReplicas returns.
> 
> For example let's say that UpdateStatefulSetNonReplicas instead of just returning `error` will return `(*string, error)`. Where that new first element would be some value we could call newImageSet with the semantics that it is set to point to appRoleImage if a container image patch was (successfully) done, and nil if no image patch was attempted or the patch failed.
> 
> If a non-nil newImageSet is returned from UpdateStatefulSetNonReplicas, then the code in the caller can do all the necessary setting of fields in the role status (UpgradingMembers and RoleUpgradeStatus).

To be honest, I'd prefer the first option with the function passing (currently is done), because `UpdateStatefulSetNonReplicas` function is more wide than application image upgrade only and we don't need to process its result after each call. But, if you insist, I will change the logic to your case.

deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
```diff
@@ -601,9 +601,8 @@ spec:
                         type: string
                       statefulSet:
                         type: string
-                      upgradingMembers:
-                        type: object
-                        nullable: true
+                      membersUpgrading:
```
joel-bluedata, line 604: Seems like we need to call this something slightly different since it is used both for upgrades and rollbacks. Hmm. I mean the obvious name is `membersUpgradingOrRollingBack` but that is a liiiiitle too long.

Let's leave it as it is for now (unless you have another great name in mind), because I'm still not sure we want to have this field in the status. I'll have another comment about that elsewhere.

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
```diff
@@ -224,12 +226,12 @@ const (
...
-	StatefulSet         string             `json:"statefulSet"`
-	Members             []MemberStatus     `json:"members"`
-	EncryptedSecretKeys map[string]string  `json:"encryptedSecretKeys,omitempty"`
-	UpgradingMembers    map[string]*string `json:"upgradingMembers,omitempty"`
-	RoleUpgradeStatus   RoleUpgradeStatus  `json:"roleUpgradeStatus,omitempty"`
+	Name                string            `json:"id"`
+	StatefulSet         string            `json:"statefulSet"`
+	Members             []MemberStatus    `json:"members"`
+	EncryptedSecretKeys map[string]string `json:"encryptedSecretKeys,omitempty"`
+	MembersUpgrading    int               `json:"membersUpgrading,omitempty"`
```
joel-bluedata, line 233: The kdcluster members field is int32, so let's make this int32 also.

joel-bluedata:
OK, I think my last concern here is in the handleCreatingMembers changes.

The func that's being modified there is one of a set of functions operating in parallel (potentially) on multiple members. So if it changes anything outside of the member-specific data, we have to worry about atomicity and races.

Also, I don't think it's valid to always decrement membersUpgrading at that spot. Technically/theoretically this could be a pod that's already been upgraded, but then got "rebooted" (deleted/recreated) for some reason while the upgrade was still ongoing for other pods. I think before decrementing you would also need to check the podUpgradeStatus to make sure this is a pod that was actually being upgraded (or rolled back). 

I have a suggestion that would address the race problem and also make it not necessary to have a membersUpgrading count in the status. Let me run through it here:

----------

First, let's not do any of this upgrade-related status management inside that asynchronous func. That way we won't even need to reason at all about atomicity for the role-status stuff. Instead, let's add more logic to the loop at the end of handleCreatingMembers. This is the way that loop currently looks:
```go
    for _, member := range creating {
        if member.State != string(memberCreating) {
            member.StateDetail.LastConfiguredContainer = member.StateDetail.ConfiguringContainer
            member.StateDetail.ConfiguringContainer = ""
        }
    }
```

This loop is looking at all the members that were processed here, and doing some state updates for any that successfully moved out of the creating state.

So we could make changes like this (untested, but you get the idea):
```go
    anyUpgradingOrRollingBack := false
    for _, member := range creating {
        if member.State != string(memberCreating) {
            member.StateDetail.LastConfiguredContainer = member.StateDetail.ConfiguringContainer
            member.StateDetail.ConfiguringContainer = ""
            member.PodUpgradeStatus = kdv1.PodConfigured
        } else {
            if member.PodUpgradeStatus != kdv1.PodConfigured {
                anyUpgradingOrRollingBack = true
            }
        }
    }
    switch (*rs).RoleUpgradeStatus {
    case kdv1.RoleUpgrading:
        if !anyUpgradingOrRollingBack {
            (*rs).RoleUpgradeStatus = kdv1.RoleUpgraded
        }
    case kdv1.RoleRollingBack:
        if !anyUpgradingOrRollingBack {
            (*rs).RoleUpgradeStatus = kdv1.RoleRolledBack
        }
    }
```

If that works, we can completely remove the membersUpgrading field.

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
joel-bluedata: (or just remove it entirely as mentioned below)

deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
Kosta91: `upgradingMembersCount` ?

pkg/apis/kubedirector/v1beta1/kubedirectorcluster_types.go
Kosta91: Currently changed to `int32`

joel-bluedata:
Gotcha. OK.

Hmm.

Well, let's stick with keeping that count in the status for now. We can still move it out of the async funcs, and add the check of podUpgradeStatus. Something like:
```go
    for _, member := range creating {
        if member.State != string(memberCreating) {
            member.StateDetail.LastConfiguredContainer = member.StateDetail.ConfiguringContainer
            member.StateDetail.ConfiguringContainer = ""
            if member.PodUpgradeStatus != kdv1.PodConfigured {
                member.PodUpgradeStatus = kdv1.PodConfigured
                (*rs).UpgradingMembersCount--
            }
        }
    }
    switch (*rs).RoleUpgradeStatus {
    case kdv1.RoleUpgrading:
        if (*rs).UpgradingMembersCount == 0 {
            (*rs).RoleUpgradeStatus = kdv1.RoleUpgraded
        }
    case kdv1.RoleRollingBack:
        if (*rs).UpgradingMembersCount == 0 {
            (*rs).RoleUpgradeStatus = kdv1.RoleRolledBack
        }
    }
```

Kosta91:
> 

Ok, this solution works well, I implemented it

joel-bluedata (approved):

PR #620, opened 2022-06-23
Quick fix

Kosta91:


joel-bluedata:
Did this problem actually happen? If so then I'm curious about how it would.

joel-bluedata:
This is OK as a safety backstop (I think), but if it can happen then we also need to fix whatever underlying issue is causing the count to be wrong.

If the count can go to 0 while some pods are still NOT in the PodConfigured state yet, then we will be transitioning the role to RoleUpgraded or RoleRolledBack state too early, which will then have other bad effects.

neogeographica (approved):

joel-bluedata (approved):

PR #621, opened 2022-07-20
rsync progress field

Kosta91:


pkg/executor/guest.go
```diff
@@ -353,7 +353,7 @@ func ExecCommand(
 	}
 
 	foundContainer := false
-	for _, containerStatus := range pod.Status.ContainerStatuses {
+	for _, containerStatus := range append(pod.Status.InitContainerStatuses, pod.Status.ContainerStatuses...) {
```
joel-bluedata, line 356: Aha, this was the problem you had earlier... good find.

pkg/executor/statefulset.go
```diff
@@ -692,7 +692,7 @@ func generateRsyncCmd(
 	// otherwise the rsync log file will not be created
 	createRsyncLogFileBaseDir := fmt.Sprintf("mkdir -p /mnt%s", filepath.Dir(kubedirectorInitLogs))
 
-	rsyncCmd := fmt.Sprintf("%s; rsync --log-file=/mnt%s --info=progress2 --relative -ax %s /mnt > /mnt%s;",
+	rsyncCmd := fmt.Sprintf("%s; rsync --log-file=/mnt%s --info=progress2 --relative -ax --no-inc-recursive %s /mnt > /mnt%s;",
```
joel-bluedata, line 695: Should we also add this argument to genrateRsyncInstalledCmd, to make sure that it is supported?

pkg/executor/cluster.go
```diff
@@ -227,3 +230,44 @@ func compactMembers(
...
+			shared.EventReasonCluster,
+			"failed to read %s",
+			progressBarFile,
+		)
+	}
+
+	if read {
+		lines := strings.Split(rsyncStatusStrB.String(), "\r")
+		lastLine := lines[len(lines)-1]
+		fields := strings.Fields(lastLine)
```
joel-bluedata, line 270: Overall this looks good, the only thing I'm really wondering about is the robustness of this rsync output parsing.

I'm wondering if maybe we should just stick the whole progress line into this property... not only would that be a little less brittle (although probably rsync does have a vested interest in keeping that progress output in a fixed format), it would also give a bit more information about things like amount copied so far, rate of copy, time spent so far, etc.

What do you think?

Kosta91: Do we really need this info out of logs? Sure, we may print the whole `lastLine value`

joel-bluedata: Yeah, I think just that whole last line. Specifically this sort of thing:
```
105.45M  13%  602.83kB/s    0:02:50 (xfr#495, ir-chk=1020/3825)
```
Showing total amount copied, and especially rate of copy and time spent so far, are at least as valuable as the current percent copied. The last bit of parenthetical stuff isn't necessarily as valuable but I don't know that we need to worry about chopping it out. I would be OK with chopping off the parenthetical portion but also OK with leaving it there.

As a bonus, if we don't bother parsing this we will be resilient to any changes in rsync progress output format.

pkg/executor/statefulset.go
Kosta91: Done

pkg/executor/cluster.go
Kosta91: Done

joel-bluedata (dismissed):

joel-bluedata:
Thought of one more thing. How does this code behave when the image doesn't have rsync, i.e. when it is using cp to do the copy and the progress file does not exist?

Kosta91:
> Thought of one more thing. How does this code behave when the image doesn't have rsync, i.e. when it is using cp to do the copy and the progress file does not exist?
> 
> (It looks like _probably_ it will fail gracefully but even so I'd rather that it not add error messages to the KD log in that case.)

Well, I added the checking if rsync is present at the container calling `rsync --version`. If it is not present, the error will be not raised or logged

joel-bluedata (approved):

PR #624, opened 2022-08-10
require golang 1.18

joel-bluedata:
Posting a draft PR of this for CI purposes. Still needs testing.

----------

Putting some CVEs in our rearview mirror.

jf docker scan before: 4 critical, 15 high
after (using golang 1.18.5): 1 critical, 8 high

The remaining critical is CVE-2022-1996 from emicklei/go-restful ... we could try to force an update on that package or we might be required to finally move to a more recent version of operator SDK. I'm not super exercised about that one though because I believe it does not apply to KD. Will double-check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - "1.16.x"
+  - "1.18.x"
 
 env:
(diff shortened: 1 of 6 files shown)


joel-bluedata:
CI seems OK. Will change to an actual PR after more regression testing.

PR #627, opened 2022-08-17
Ezml 1747 correct rollback process

Kosta91:
Made rollback process more stable.
Eliminated UpgradeMembersCount field

### DIFF
--- a/pkg/controller/kubedirectorcluster/members.go
+++ b/pkg/controller/kubedirectorcluster/members.go
@@ -424,6 +424,34 @@ func handleReadyMembers(
 
 			m.StateDetail.LastConfigDataGeneration = cr.Status.SpecGenerationToProcess
 		}(member)
+
+		updateRoleUpgradeStatus := func() {
+
+			rs := &role.roleStatus
+			switch (*rs).RoleUpgradeStatus {
+			case kdv1.RoleUpgrading:
+				for _, member := range role.roleStatus.Members {
+					if member.PodUpgradeStatus != kdv1.PodUpgraded {
+						return
+					}
+				}
+				(*rs).RoleUpgradeStatus = kdv1.RoleUpgraded
+				reqLogger.Info(fmt.Sprintf("role %s -> %s", role.roleSpec.Name, (*rs).RoleUpgradeStatus))
+
+			case kdv1.RoleRollingBack:
+				for _, member := range role.roleStatus.Members {
+					if member.PodUpgradeStatus != kdv1.PodRolledBack {
+						return
+					}
+				}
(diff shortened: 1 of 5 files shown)


pkg/controller/kubedirectorcluster/members.go
```diff
@@ -424,6 +424,34 @@ func handleReadyMembers(
 
 			m.StateDetail.LastConfigDataGeneration = cr.Status.SpecGenerationToProcess
 		}(member)
+
+		updateRoleUpgradeStatus := func() {
```
joel-bluedata, line 428: I'm dubious that this is the right place to run this code, for a couple of reasons.

First thing: You currently have it inside the loop that is iterating over every ready member in the role. So if there are for example three ready members then updateRoleUpgradeStatus will be run three times in this loop, which seems overkill since each call of updateRoleUpgradeStatus iterates over all the role members.

So maybe one thing is that you can/should just pop it outside of this ready-members loop. However I think that still leaves the question of whether we should be doing this role status management inside handleReadyMembers at all. (It doesn't seem to be related to the "mission" of updating ready-status members with new configmeta.) You're setting the final pod upgrade status as part of handleCreatingMembers... maybe setting the role upgrade status should go there too?

Kosta91: Got it, I moved the logic to `syncRoles` method.

pkg/controller/kubedirectorcluster/roles.go
```diff
@@ -120,6 +124,32 @@ func syncClusterRoles(
 	return roles, returnState, nil
 }
 
+func updateRoleUpgradeStatus(
```
joel-bluedata, line 127: Needs a function header comment here.

pkg/controller/kubedirectorcluster/roles.go
```diff
@@ -120,6 +124,32 @@ func syncClusterRoles(
...
+	rs *kdv1.RoleStatus,
+) {
+	if rs == nil {
```
joel-bluedata, line 130: blank line for code convention

joel-bluedata:
Looks good I think, just a couple of code convention things.

pkg/controller/kubedirectorcluster/roles.go
Kosta91: Done

pkg/controller/kubedirectorcluster/roles.go
Kosta91: Done

joel-bluedata (approved):

Issue #601, opened 2022-05-02
eventlist support for all startscript args


joel-bluedata:
I need to make sure that our new upgrade-related startscript notifications are included in kdapp eventlist specifications, i.e. an app can opt out of receiving them.

In the process I also need to go back and look at the reconnect notification; I don't think eventlist support was added for that at the time it was implemented. We can bring it into the fold now but would need to have a bit of a think about how that would affect existing apps that enumerate an eventlist.

Issue #402, opened 2020-08-28
/bin/sh kd-JSON file config


hoycdanny:
Hi Team, 
How can I adjust the kd-JSON file config, if my Image does not have bash but only sh?

Error Message
`OCI runtime exec failed: exec failed: container_linux.go:349: starting container process caused "exec: \"/bin/bash\": stat /bin/bash: no such file or directory": unknown`

```
{
    "apiVersion": "kubedirector.hpe.com/v1beta1",
    "kind": "KubeDirectorApp",
    "metadata": {
        "name" : "httpd"
    },

    "spec" : {
        "systemdRequired": true,
        "defaultPersistDirs" : ["/usr/local/apache2/htdocs"],
        "config": {
            "roleServices": [
                {
                    "serviceIDs": [
                        "http"
                    ],
                    "roleID": "httpd"
                }
            ],
            "selectedRoles": [
                "httpd"
            ]
        },
        "label": {
            "name": "HTTPD",
            "description": "Httpd apps"
        },
        "distroID": "apache/httpd",
        "version": "2.4.46-alpine",
        "configSchemaVersion": 8,
        "services": [
            {
                "endpoint": {
                    "urlScheme": "http",
                    "path": "/",
                    "isDashboard": true,
                    "port": 80
                },
                "id": "http",
                "label": {
                    "name": "http"
                }
            }
        ],
        "defaultImageRepoTag": "httpd:2.4.46-alpine",
        "defaultConfigPackage": null,
        "roles": [
            {
                "cardinality": "1+",
                "id": "httpd"
            }
        ]
    }
}
```

joel-bluedata:
Hmm. bash is a requirement at the moment. I'm not sure though if any of the current shell executions rely on bash features, so we may be able to change that.

hoycdanny:
But some light images only provide /bin/sh, for example apache https and busybody.

joel-bluedata:
Yup very true. I think I'm not clear on why you would need/want to use KubeDirector for those particular examples, but I can easily imagine there being other cases where bash would not be available.

hoycdanny:
Because I hope that customers can deploy the service through HPECP kd dashboard one click, our current deployment service dashboard through kubectl YAML is not friendly.

joel-bluedata:
I'm back from leave; will take this up. Note to self... the wiki currently (now) lists bash as an image requirement, so I would need to revert that if we use sh instead.

joel-bluedata:
[...]

PR #633, opened 2022-08-24
Backup and restore /opt/guestconfig folder

Kosta91:


pkg/controller/kubedirectorcluster/members.go
```diff
@@ -577,6 +667,8 @@ func handleCreatingMembers(
 				}
 			}
 
+			handleConfigPackage(reqLogger, cr, m)
```
joel-bluedata, line 670: There's a few error cases in handleConfigPackage that we need to decide how to handle (rather than only logging the error).

If we return early out of handleConfigPackage, returning an error, then we have two choices of what to do here in that case:

* We could return from here WITHOUT doing setFinalState. This would leave the member in creating state and it will be processed again on next reconciliation.

* Or we could use setFinalState to put the member into config error, and return.

Depends on if we think the possible errors causing this are transient or not. At first glance, seems like they are not transient, so we should probably just put the member into config error state.

(If they happen to be caused by the pod going away, then when the pod comes back we'll go through create-pending and creating states again and try everything again.)

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -509,6 +510,95 @@ func handleCreatePendingMembers(
...
+	if podErr != nil {
+		logErrWrapper(podErr, fmt.Sprintf("failed to get pod %s", member.Pod))
+	}
+
+	execCommandWrapper := func(cmd string) error {
+
+		var stdOut bytes.Buffer
+		ioStreams := &executor.Streams{Out: &stdOut}
+
+		return executor.ExecCommand(
```
joel-bluedata, line 545: This is a fine way to run the command I think, but it also seems pretty similar to what executor.RunScript does... I'm curious if you tried that and there was a problem with it.

BTW if we do want/need to do it this way instead of using RunScript, then we should probably export executor.execShell (i.e. turn it into executor.ExecShell) and use that here instead of hardcoding "/bin/bash". I realize we do have /bin/bash hardcoded in a couple of other places too, but we'll need to deal with all of those if/when we tackle issue #402 .

pkg/executor/guest.go
```diff
@@ -377,6 +382,7 @@ func ExecCommand(
 			pod.Status.Phase,
 		)
 	}
+	reqLogger.Info(fmt.Sprint("executing command: ", strings.Join(command, " ")))
```
joel-bluedata, line 385: debug print that we can remove?

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -1466,8 +1561,25 @@ func appConfig(
 	if role.EventList != nil && !shared.StringInList("configure", *role.EventList) {
 		return true, nil
 	}
+
+	// Notify upgrade/rollback completion as necessary.
```
joel-bluedata, line 1565: Didn't we already send these notifications above?

joel-bluedata:
It seems like we would need to add appConfigBackupPath to the list of appConfigDefaultMountFolders to make sure that it gets on persistent storage, otherwise it would be lost when the pod reboots.

pkg/controller/kubedirectorcluster/members.go
joel-bluedata: Ah, the code above probably doesn't get triggered any longer because we moved away the directory that contains the config status file. I think we can delete that older code that was intended to send the upgraded/reverted notifications.

pkg/controller/kubedirectorcluster/members.go
Kosta91: Yes, `RunScript` works ok, fixed

pkg/executor/guest.go
Kosta91: Removed

pkg/controller/kubedirectorcluster/members.go
Kosta91: Ok, fixed

Kosta91:
> It seems like we would need to add appConfigBackupPath to the list of appConfigDefaultMountFolders to make sure that it gets on persistent storage, otherwise it would be lost when the pod reboots.

Done

pkg/controller/kubedirectorcluster/members.go
```diff
@@ -577,6 +651,16 @@ func handleCreatingMembers(
...
 
+			backupErr := handleConfigPackage(reqLogger, cr, m)
+
+			if backupErr != nil {
+				statusErrMsg := fmt.Sprintf(
+					"failed to backup config package after upgrade: %s",
+					backupErr.Error(),
+				)
+				setFinalState(memberConfigError, &statusErrMsg)
+			}
```
joel-bluedata, line 662: We would need to `return` here in the error case, correct?

Kosta91: Sure

pkg/controller/kubedirectorcluster/members.go
joel-bluedata: Probably need to move this notification code up above the "Run the config file iff ..." comment. That code block above is an early bailout of this function if the "configure" event is not desired for this app. I guess I can imagine situations where the "configure" event is not desired but the upgraded/rollback events are.

(We should add eventlist checking for the upgraded/rolledback events too, as per issue #601, but that can be a separate PR.)

Kosta91: Done

joel-bluedata (approved):

PR #634, opened 2022-09-22
Ezml 1809 upgrade notifications on cluster/role levels

Kosta91:


neogeographica:
This is really good! The only thing I'm still dubious about here is the error case in RunConfigScript when a pod member is down. As the code currently stands the pod will lose that notification forever, in the case of role and cluster level notifications.

For those role/cluster notifications we may need to make use of the "notification queueing" mechanism to ensure that those notifications don't get lost. I.e. instead of immediately sending them to to pod, instead we can append them to the member's PendingNotifyCmds. You can have a grep in the code to see how those are handled, but the gist is that currently notifications are placed on this queue using queueNotify in members.go, and then the queued notifies are processed by syncMemberNotifies in members.go. syncMemberNotifies is called on every cluster reconciliation pass, so if a particular notification doesn't get successfully taken off the queue (because a member is down) it will get retried later.

A few considerations if we use that mechanism:
* Currently queueNotify assumes that it is only used for addnodes/delnodes events. To make it useful for other kinds of events it would need to modified (or, quite possibly, a different similar function created -- whichever is cleaner).
* syncMemberNotifies currently takes care of advancing LastSetupGeneration to match LastConfigDataGeneration when it sends a member a notify, because up until now it is dealing with notifies that are reactions to cluster reconfigurations. In this case that should be a NOP, i.e. the LastSetupGeneration==LastConfigDataGeneration already, so it's probably fine.
* I'm wondering how this affects the state cleanup in handleClusterUpgrade. It does seem like we should _not_ finish the upgrade/rollback - i.e. removing the app reference and clearing the upgradeinfo -- until we know all the notifications have been sent (all the notify queues are empty). Otherwise someone could start reconfiguring the cluster while there are still upgrade notifies that need to be sent.

Let me know your thoughts about all that.

joel-bluedata:
This is really good! The only thing I'm still dubious about here is the error case in RunConfigScript when a pod member is down. As the code currently stands the pod will lose that notification forever, in the case of role and cluster level notifications.

For those role/cluster notifications we may need to make use of the "notification queueing" mechanism to ensure that those notifications don't get lost. I.e. instead of immediately sending them to to pod, instead we can append them to the member's PendingNotifyCmds. You can have a grep in the code to see how those are handled, but the gist is that currently notifications are placed on this queue using queueNotify in members.go, and then the queued notifies are processed by syncMemberNotifies in members.go. syncMemberNotifies is called on every cluster reconciliation pass, so if a particular notification doesn't get successfully taken off the queue (because a member is down) it will get retried later.

A few considerations if we use that mechanism:
* Currently queueNotify assumes that it is only used for addnodes/delnodes events. To make it useful for other kinds of events it would need to modified (or, quite possibly, a different similar function created -- whichever is cleaner).
* syncMemberNotifies currently takes care of advancing LastSetupGeneration to match LastConfigDataGeneration when it sends a member a notify, because up until now it is dealing with notifies that are reactions to cluster reconfigurations. In this case that should be a NOP, i.e. the LastSetupGeneration==LastConfigDataGeneration already, so it's probably fine.
* I'm wondering how this affects the state cleanup in handleClusterUpgrade. It does seem like we should _not_ finish the upgrade/rollback - i.e. removing the app reference and clearing the upgradeinfo -- until we know all the notifications have been sent (all the notify queues are empty). Otherwise someone could start reconfiguring the cluster while there are still upgrade notifies that need to be sent.

Let me know your thoughts about all that.

joel-bluedata:
You mentioned that queued notifies weren't getting sent.

I think you're being bitten by this problem: https://www.darrenlester.com/blog/go-range-gotcha

At the end of updateRoleUpgradeStatus for example, and in handleClusterUpgrade. Specifically when you get "member" from the range and then modify some field of member, you're not really modifying the original member object.

Kosta91:
> You mentioned that queued notifies weren't getting sent.
> 
> I think you're being bitten by this problem: https://www.darrenlester.com/blog/go-range-gotcha
> 
> At the end of updateRoleUpgradeStatus for example, and in handleClusterUpgrade. Specifically when you get "member" from the range and then modify some field of member, you're not really modifying the original member object.

Sure, I already got it)

pkg/controller/kubedirectorcluster/util.go
```diff
@@ -35,3 +41,147 @@ func updateSchedulingErrorMessage(
...
+		"--role",
+		roleName,
+		"--fqdns",
+		deltaFqdns,
+	}
+	notifyDesc := kdv1.NotificationDesc{
+		Arguments: arguments,
+	}
+
+	for i, role := range cr.Status.Roles {
```
joel-bluedata, line 141: This looping in QueueNotify will work, but it makes me sad. :-)

I think we can avoid this by passing in a pointer to the member. That means that in the caller, instead of using `range` to iterate over the members, it will have to iterate over the count of members (let's say that is `j`) and get each member by its array subscript (e.g. `Members[j]`) ... then the caller will be working with the actual member object (not a copy) and can pass its pointer to QueueNotify.

There's various examples of that approach in the code. `range` is a real troublemaker.

Kosta91: > This looping in QueueNotify will work, but it makes me sad. :-)
> 
> I think we can avoid this by passing in a pointer to the member. That means that in the caller, instead of using `range` to iterate over the members, it will have to iterate over the count of members (let's say that is `j`) and get each member by its array subscript (e.g. `Members[j]`) ... then the caller will be working with the actual member object (not a copy) and can pass its pointer to QueueNotify.
> 
> There's various examples of that approach in the code. `range` is a real troublemaker.

Well, if I understood correctly, I replaced it to `for j := 0; j < len(role.Members); j++ {...}` loop form

neogeographica:
You still have that loop at the end of QueueNotify though. I think the start of this discussion was how to avoid doing that loop by passing in the member state detail as an argument.

joel-bluedata:
You still have that loop at the end of QueueNotify though. I think the start of this discussion was how to avoid doing that loop by passing in the member state detail as an argument.

Kosta91:
> You still have that loop at the end of QueueNotify though. I think the start of this discussion was how to avoid doing that loop by passing in the member state detail as an argument.

@joel-bluedata , fixed

joel-bluedata (approved):

Issue #618, opened 2022-06-09
make sure non-image-changing upgrades work


joel-bluedata:
Part of issue https://github.com/bluek8s/kubedirector/issues/588 ; related to issue #617.

It's quite possible that an upgrade will change some things about the kdapp (including the config package) but NOT change the container image. In which case KD may have things it needs to do to the container (e.g. issue #617 for config package changes) and we need to make sure that it is not completely relying on container-image-change as an indicator of an upgrade.

joel-bluedata:
For the MVP I think it would be completely acceptable for the validation code to enforce that, when switching kdapps, the image MUST change.

PR #636, opened 2022-11-07
Stricted validating application changes for live upgrade

Kosta91:


Kosta91:
https://github.com/bluek8s/kubedirector/issues/618

pkg/validator/cluster.go
```diff
@@ -461,12 +461,30 @@ func validateAppSpecChanges(
...
-				(&prevApp.Spec.NodeRoles[prevAppRoleIdx].SetupPackage.Info).PackageURL = appRole.SetupPackage.Info.PackageURL
+				if *prevAppRole.ImageRepoTag != *appRole.ImageRepoTag {
+					prevAppRole.ImageRepoTag = appRole.ImageRepoTag
+					roleImageChanged = true
+				}
+
+				if prevAppRole.SetupPackage.Info.PackageURL != appRole.SetupPackage.Info.PackageURL {
+					// https://github.com/bluek8s/kubedirector/issues/618
+					// If the role image was not changed but some other role attributes were changed
+					// decline the upgrade
```
joel-bluedata, line 478: Took me a bit to get on board with this, but yeah I think this is correct.

Let me talk through how I see it.

Recap of the requirement: if the kdapp spec for a kdcluster was changed -- which must be the case since we're in this function -- then we want the current containers in a given role to either be completely unaffected by the switch between the new and old kdapp, or the role must change its image.

We currently only allow two changes in the kdapp: package URL and cardinality. Cardinality doesn't affect the current containers. If package URL is changing, we make sure the image is changing too.

Seems OK!

joel-bluedata (approved):

PR #637, opened 2022-11-21
support custom size for /dev/shm

joel-bluedata:
The memory backing /dev/shm in a container is normally limited to 64MB. Some applications need (a lot) more than that, e.g. PyTorch jobs.

At a glance it seems like a solution would be to mount an emptyDir memory volume to /dev/shm. However until (somewhat) recently you couldn't specify the max size of such a volume, and it would be half of your host's RAM. That's ... not good.

K8s 1.20 add a feature flag (false by default) to enable setting a size limit on such volumes. With K8s 1.22, that feature flag is true by default. If all goes well it should eventually graduate out of feature-flag-land but hasn't yet.

=====

This change introduces an optional quantity for role spec in a kdcluster: "sharedMemory". So for example your kdcluster role spec could include
```yaml
sharedMemory: "1Gi"
```

We've been running test builds of this for quite a while with a user that has a demanding PyTorch setup and this seems to work great!

The problem is, how do you deal with K8s clusters that don't know about this setting? If you try to do this and your K8s cluster doesn't have the SizeMemoryBackedVolumes feature gate enabled, then you'll end up with the ginormous memory volume as mentioned above.

So...

By default, KD will assume that you are allowed to specify sharedMemory in a role spec if your K8s server is version 1.22 or better.

If you want to change this behavior -- let's say you've explicitly disabled the feature gate in your K8s 1.22+ cluster, or maybe you enabled the feature gate in an earlier K8s version -- you can indicate that with an optional boolean flag in the KD config: "forceSharedMemorySizeSupport".

If you leave that attribute unspecified in the KD config, you'll get the version-based gating described above, i.e. you can use sharedMemory in your role specs if your K8s cluster is 1.22+. However you can also choose to explicitly set this attribute to a value of true or false to suppress the version-based check in favor of your explicit declaration.

=====

Also included: a minor improvement to validateRoleStorageClass. If parsing a storage quantity in a role fails, it will continue to check the other roles rather than bailing immediately. If you have any similar issues in other roles this will let you see them all at once.

### DIFF
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorclusters_crd.yaml
@@ -197,6 +197,9 @@ spec:
                               minLength: 1
                             value:
                               type: string
+                      sharedMemory:
+                        type: string
+                        pattern: '^([0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
                       storage:
                         type: object
                         nullable: true
--- a/deploy/kubedirector/kubedirector.hpe.com_kubedirectorconfigs_crd.yaml
+++ b/deploy/kubedirector/kubedirector.hpe.com_kubedirectorconfigs_crd.yaml
@@ -78,6 +78,8 @@ spec:
                   type: boolean
                 allowRestoreWithoutConnections:
                   type: boolean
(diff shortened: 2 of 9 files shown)


joel-bluedata:
(Ran out of Travis CI credits for November apparently.)

Kosta91 (approved):

PR #658, opened 2026-03-25
Qwen2 support for kubedirector

vickyuec:


deploy/example_clusters/cr-cluster-qwen2-test.yaml
```diff
@@ -0,0 +1,23 @@
...
+  name: qwen2-test-cluster
+spec:
+  app: qwen2-server
+  appCatalog: local
+  roles:
+  - id: inference-server
+    members: 1
+    resources:
+      requests:
+        memory: "1Gi"                # Qwen2 requires more memory
```
kmathur, line 13: if this role will run the inference server we should increase memory

deploy/example_catalog/cr-app-qwen2.yaml
```diff
@@ -0,0 +1,43 @@
...
+  label:
+    name: "Qwen2 Inference Server"
+    description: "Qwen2 large language model inference server powered by Ollama"
+  distroID: "qwen2/server"
+  version: "1.0"
+  configSchemaVersion: 8
+  systemdRequired: false
+  
+  # Use Ollama for Qwen2 model inference
+  defaultImageRepoTag: "ollama/ollama:latest"
```
kmathur, line 15: at what point is the qwen model downloaded ? for future we should create a new image with qwen2 and ollama

kmathur:
left some suggestions, overall looks good. This will be a great new llm model in the kubedirector ecosystem

